PR #5, opened 2018-11-15
Add whitelisting of files and folders, copy files from src to lib

tbergq:
Move dependencies to devDependencies

### DIFF
--- a/package.json
+++ b/package.json
@@ -2,7 +2,7 @@
   "name": "universal-orbit-components",
   "version": "0.0.1",
   "description": "The proof of concept of universal components.",
-  "main": "index.js",
+  "main": "lib/index.js",
   "scripts": {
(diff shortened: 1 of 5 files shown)


package.json
```diff
@@ -23,31 +29,38 @@
   "author": "kiwi.com",
   "license": "MIT",
   "dependencies": {
+
+  },
+  "peerDependencies": {
     "react": "16.6.1",
-    "react-art": "16.6.1",
-    "react-dom": "16.6.1",
-    "react-native": "^0.57.5",
```
tbergq: having this in dependencies will cause crash for consumer, saying duplicate import of react-native

FilipMessa: Thank with these libraries I wasn't been sure where it belongs

PR #7, opened 2018-11-15
Add Android on README and remove .eslintignore

ferrannp:
- No need of .eslintignore and I think the less files the better


package.json
```diff
@@ -10,7 +10,7 @@
     "storybook:build": "build-storybook -c .storybook",
     "prestorybook-native": "node ./node_modules/.bin/rnstl",
     "storybook-native": "storybook -c storybook.native start -p 7007",
-    "eslint": "eslint .",
```
tbergq: Why this change? There could be files outside src folder we want to lint

PR #8, opened 2018-11-15
Add Flow

### LINKED ISSUE #2: Add flow

ferrannp:
Closes #2.

- Uses Flow that comes with latest `react-native init`

### DIFF
--- a/.flowconfig
+++ b/.flowconfig
@@ -69,5 +69,8 @@ suppress_comment=\\(.\\|\n\\)*\\$FlowIssue\\((\\(<VERSION>\\)? *\\(site=[a-z,_]*
 suppress_comment=\\(.\\|\n\\)*\\$FlowFixedInNextDeploy
 suppress_comment=\\(.\\|\n\\)*\\$FlowExpectedError
 
+[lints]
+unnecessary-optional-chain=error
+
 [version]
 ^0.78.0
--- a/.flowconfig
+++ b/.flowconfig
@@ -1,0 +1,73 @@
+[ignore]
+; We fork some components by platform
+.*/*[.]android.js
+
+; Ignore "BUCK" generated dirs
+<PROJECT_ROOT>/\.buckd/
+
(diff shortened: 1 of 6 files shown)


.flowconfig
```diff
@@ -0,0 +1,73 @@
...
+suppress_type=$FlowFixMeProps
+suppress_type=$FlowFixMeState
+
+suppress_comment=\\(.\\|\n\\)*\\$FlowFixMe\\($\\|[^(]\\|(\\(<VERSION>\\)? *\\(site=[a-z,_]*react_native[a-z,_]*\\)?)\\)
+suppress_comment=\\(.\\|\n\\)*\\$FlowIssue\\((\\(<VERSION>\\)? *\\(site=[a-z,_]*react_native[a-z,_]*\\)?)\\)?:? #[0-9]+
+suppress_comment=\\(.\\|\n\\)*\\$FlowFixedInNextDeploy
+suppress_comment=\\(.\\|\n\\)*\\$FlowExpectedError
+
+[version]
+^0.78.0
```
tbergq: can you add latest version of flow?

.flowconfig
```diff
@@ -0,0 +1,73 @@
...
+[libs]
+node_modules/react-native/Libraries/react-native/react-native-interface.js
+node_modules/react-native/flow/
+node_modules/react-native/flow-github/
+
+[options]
+emoji=true
+
+esproposal.optional_chaining=enable
+esproposal.nullish_coalescing=enable
```
tbergq: could you also add 

```
[lints]
unnecessary-optional-chain=error
```

.flowconfig
ferrannp: Not really, it gives several errors with RN. If you check my comment, I added the one that comes with `react-native init`. We should keep with that and update it when they update it IMO.

tbergq: I prefer to be up to date and rather add the filepaths that gives errors to [untyped] but, I will allow it 🤔

PR #10, opened 2018-11-15
Add eslint plugin flow

tbergq:


ferrannp:
@tbergq would be nice you do some PRs into https://github.com/callstack/eslint-config-callstack/ :)

PR #14, opened 2018-11-15
Add script in package.json for publishing

tbergq:


package.json
```diff
@@ -14,7 +14,8 @@
     "flow": "flow",
     "rnstl": "rnstl",
     "ios": "rnstl && react-native run-ios",
-    "android": "rnstl && react-native run-android"
+    "android": "rnstl && react-native run-android",
+    "npm-publish": "yarn babel-node scripts/createPackage.js && yarn pack"
```
tbergq: `yarn pack` should be replaced with `yarn publish` when we are ready to publish, but for now this is ok, so we can test it locally in mobile repo

PR #20, opened 2018-11-15
Add Icon component

ferrannp:
- It's still missing to decide exactly how to handle it on web


tbergq:
image

src/Icon/index.js
```diff
@@ -0,0 +1,45 @@
...
+  }
+  return iconsMap[name].character;
+};
+
+const Icon = ({ name, color, size }: Props) => (
+  <RNText
+    color={color}
+    size={size}
+    style={[styles.icon, { color: color || '#46515e', fontSize: size || 20 }]}
+  >
```
tbergq: `{ color: color ?? '#46515e'`
`fontSize: size ?? 20` -> since passing fontSize 0 would fallback to 20, which is not what you expect

src/index.js
```diff
@@ -2,6 +2,7 @@
 
 /* Components */
 export { default as Text } from './Text';
+export { default as Icon } from './Icon';
```
tbergq: I don't think we should export icons from this package.

tbergq:
Also, maybe it should go into its own folder, and use yarn workspaces?

ferrannp:
@tbergq you have red screen because you  need to use `yarn link` so it won't look for this package in `node_modules`. I will look into workspaces.

tbergq:
I just added it to a brand new react-native project and did react-native link @kiwicom/react-native-orbit-icons

Consider this a package that anyone should be able to use. 

If there are additional steps, please add them to a readme file

ferrannp:
> Consider this a package that anyone should be able to use.

That will be when it's in npm.

> If there are additional steps, please add them to a readme file

Let me check workspaces first.

tbergq:
> That will be when it's in npm.

Yes, but still, I expect it to work like that before it is merged.

I did npm pack in this repo, and added it to my rntesting project using the file path. Which is the same as installing it from npm

src/Icon/index.js
```diff
@@ -0,0 +1,45 @@
...
+
+const Icon = ({ name, color, size }: Props) => (
+  <RNText
+    color={color}
+    size={size}
+    style={[styles.icon, { color: color ?? '#46515e', fontSize: size ?? 20 }]}
+  >
+    {getIconCharacter(name)}
+  </RNText>
+);
```
tbergq: Just realised, how about

```
Icon.defaultProps = {
  color: '#46515e',
  size: 20,
}
```

then just 

```
const Icon = ({ name, color, size }: Props) => (
  <RNText
    color={color}
    size={size}
    style={[styles.icon, { color, fontSize: size }]}
  >´´´

PR #22, opened 2018-11-16
TextInput

lukewalczak:
- [x] - default input
- [x] - disabled input
- [x] - small input
- [x] - with text prefix
- [x] - compact input
- [x] - required field
- [x] - with Icon prefix

AFTER THAT PR OPEN ISSUES FOR:

1. TextInput test
2. TextInput suffix - enhancement
3. TextInput types (number, password, email) - enhancement
4. TextInput helper - enhancement

### DIFF
--- a/src/TextInput/FormLabel.js
+++ b/src/TextInput/FormLabel.js
@@ -7,7 +7,7 @@ import { View, Text } from 'react-native';
 import { StyleSheet } from '..';
 
 type Props = {|
-  +children: React$Node,
+  +children: React.Node,
   +filled?: boolean,
   +disabled?: boolean,
   +required?: boolean,
@@ -16,10 +16,10 @@ type Props = {|
 
 const getAsteriksStyle = (filled, disabled) => {
   if (disabled) {
-    return styles.asteriksFilled;
+    return styles.asteriksDisabled;
   }
   if (filled) {
-    return styles.asteriksDisabled;
+    return styles.asteriksFilled;
   }
   return styles.asteriksDefault;
 };
--- a/src/TextInput/index.js
+++ b/src/TextInput/index.js
@@ -15,7 +15,7 @@ type Props = {|
   +required?: boolean,
   +inlineLabel?: boolean,
   +label: string,
-  +prefix?: React$Node,
+  +prefix?: React.Node,
   +onFocus?: () => void | Promise<any>,
   +onBlur?: () => void | Promise<any>,
   +onChange?: () => void | Promise<any>,


src/TextInput/FormLabel.js
```diff
@@ -0,0 +1,53 @@
+// @flow
+
+import * as React from 'react';
+import { View, Text } from 'react-native';
+
+import { StyleSheet } from '..';
+
+type Props = {
```
tbergq: should be exact

tbergq: and readonly

src/TextInput/index.js
```diff
@@ -0,0 +1,118 @@
+// @flow
+
+import * as React from 'react';
+import { TextInput as RNTextInput, View } from 'react-native';
+
+import { StyleSheet } from '..';
+import FormLabel from './FormLabel';
+
+type Props = {
```
tbergq: should be exact

src/TextInput/index.js
```diff
@@ -0,0 +1,118 @@
...
+
+const colors = {
+  colorTextInputDisabled: '#bac7d5',
+  colorTextInput: '#46515e',
+  borderColorInput: '#bac7d5',
+  colorPlaceholderInput: '#bac7d5',
+  borderColorInputFocus: '#0176D2',
+};
+
+const getFontSize = size => (size === 'small' ? 14 : 14);
```
tbergq: why is it the same if size is small, or not small?

src/TextInput/index.js
```diff
@@ -0,0 +1,118 @@
...
+type State = {
+  isFocused: boolean,
+};
+
+const colors = {
+  colorTextInputDisabled: '#bac7d5',
+  colorTextInput: '#46515e',
+  borderColorInput: '#bac7d5',
+  colorPlaceholderInput: '#bac7d5',
+  borderColorInputFocus: '#0176D2',
```
tbergq: Should we not get this colors from orbit design tokens?

src/TextInput/index.js
```diff
@@ -0,0 +1,118 @@
...
+import FormLabel from './FormLabel';
+
+type Props = {
+  +size?: 'small' | 'normal',
+  +placeholder?: string,
+  +value?: string,
+  +disabled?: boolean,
+  +onFocus?: () => void | Promise<any>,
+  +onBlur?: () => void | Promise<any>,
+  required?: boolean,
```
tbergq: should also be readonly

src/TextInput/index.js
lukewalczak: Just take these values from https://github.com/kiwicom/orbit-components

src/TextInput/index.js
```diff
@@ -0,0 +1,118 @@
...
+  +label: string,
+};
+
+type State = {
+  isFocused: boolean,
```
tbergq: please use exact types

src/TextInput/index.js
```diff
@@ -0,0 +1,118 @@
...
+  colorTextInputDisabled: '#bac7d5',
+  colorTextInput: '#46515e',
+  borderColorInput: '#bac7d5',
+  colorPlaceholderInput: '#bac7d5',
+  borderColorInputFocus: '#0176D2',
+};
+
+const getFontSize = size => (size === 'small' ? 14 : 14);
+const getHeight = size => (size === 'small' ? 32 : 44);
+const getPadding = size => (size === 'small' ? 12 : 12);
```
tbergq: also here, why the same value return?

src/TextInput/index.js
lukewalczak: At the end of the work on `TextInput` I'll take them from design tokens

src/TextInput/index.js
```diff
@@ -0,0 +1,118 @@
...
+            {
+              height: getHeight(size),
+              paddingHorizontal: getPadding(size),
+              lineHeight: getHeight(size),
+              borderColor: getBorderColor(isFocused),
+            },
+          ]}
+        >
+          <RNTextInput
+            onFocus={() => {
```
tbergq: please don't use arrow functions

src/TextInput/index.js
tbergq: I think we should rather use these `import { defaultTokens } from '@kiwicom/mobile-orbit';`

lukewalczak: https://github.com/kiwicom/orbit-design-tokens/blob/4930822a7ae11dca6026d5845d37d6eb65b00907/output/theo-spec.common.js#L27

https://github.com/kiwicom/orbit-design-tokens/blob/4930822a7ae11dca6026d5845d37d6eb65b00907/output/theo-spec.common.js#L166

tbergq: That is ok, but then is function is useless, just set fontSize to 14

src/TextInput/FormLabel.js
```diff
@@ -0,0 +1,54 @@
...
+
+const Asteriks = ({ filled, children }) => (
+  <Text style={[styles.asteriks, { color: getAsteriksColor(filled) }]}>
+    {children}
+  </Text>
+);
+
+const FormLabel = ({ children, required, filled, disabled }: Props) => (
+  <View style={[styles.formLabel, { color: getColor(filled, disabled) }]}>
+    {required && <Asteriks filled={filled}>* </Asteriks>}
```
tbergq: Instaed of doing 
```
<View style={[styles.formLabel, { color: getColor(filled, disabled) }]}>
```
 since there are only two options, could you just create the styles with StyleSheet.create, and select the correct style?

src/TextInput/FormLabel.js
```diff
@@ -0,0 +1,69 @@
...
+    ? defaultTokens.orbit.colorTextError
+    : defaultTokens.orbit.colorFormLabelFilled;
+
+const getColor = (filled, disabled) =>
+  !filled || disabled
+    ? defaultTokens.orbit.colorFormLabel
+    : defaultTokens.orbit.colorFormLabelFilled;
+
+const Asteriks = ({ filled, children }) => (
+  <Text style={[styles.asteriks, { color: getAsteriksColor(filled) }]}>
```
RobinCsl: I believe that means that this style will not take advantage of `StyleSheet.create`. What do you think about my approach there: https://github.com/kiwicom/universal-orbit-components/blob/7dd2afa002f089c4a898ce02f00be0147bf2687c/src/Button/ButtonTitle.js#L32

src/TextInput/FormLabel.js
```diff
@@ -0,0 +1,69 @@
...
+  children,
+  required,
+  filled,
+  disabled,
+  inlineLabel,
+}: Props) => (
+  <View style={inlineLabel ? styles.inlineFormLabel : styles.formLabel}>
+    {required && (
+      <Asteriks style={{ color: getColor(filled, disabled) }} filled={filled}>
+        *{' '}
```
RobinCsl: This might break with RTL support, but I am not sure what the best way to do it would be. 🤔

src/TextInput/index.js
```diff
@@ -0,0 +1,176 @@
...
+  +required?: boolean,
+  +inlineLabel?: boolean,
+  +label: string,
+  +prefix?: React$Node,
+  +onFocus?: () => void | Promise<any>,
+  +onBlur?: () => void | Promise<any>,
+  +onChange?: () => void | Promise<any>,
+|};
+
+type State = {
```
RobinCsl:
```suggestion
type State = {|
```

src/TextInput/index.js
```diff
@@ -0,0 +1,176 @@
...
+  isFocused: boolean,
+};
```
RobinCsl:
```suggestion
|};
```

src/TextInput/FormLabel.js
```diff
@@ -0,0 +1,69 @@
...
+const FormLabel = ({
+  children,
+  required,
+  filled,
+  disabled,
+  inlineLabel,
+}: Props) => (
+  <View style={inlineLabel ? styles.inlineFormLabel : styles.formLabel}>
+    {required && (
+      <Asteriks style={{ color: getColor(filled, disabled) }} filled={filled}>
```
tbergq: I would prefer that these ternaries pick some cached style instead, so getColor could instead just do:

```
const getColor = filled => filled ? return styles.filled : styles.notFilled;
```

src/TextInput/index.js
tbergq: This type is still not exact

src/TextInput/index.js
```diff
@@ -0,0 +1,176 @@
...
+    ? defaultTokens.orbit.colorTextInputDisabled
+    : defaultTokens.orbit.colorTextInput;
+const getBorderColor = isFocused =>
+  isFocused
+    ? defaultTokens.orbit.borderColorInputFocus
+    : defaultTokens.orbit.borderColorInput;
+
+const Prefix = ({ children, size }) => {
+  if (typeof children === 'object' && children.props) {
+    const icon = {
```
tbergq: React.cloneElement?

src/TextInput/index.js
```diff
@@ -0,0 +1,176 @@
...
+          <FormLabel filled={!!value} required={required}>
+            {label}
+          </FormLabel>
+        )}
+        <View
+          style={[
+            styles.inputContainer,
+            {
+              height: getHeight(size),
+              borderColor: getBorderColor(isFocused),
```
tbergq: this should also fetch from cached styles

Issue #18, opened 2018-11-15
Create the Button component


RobinCsl:
Tasks to enhance the Button component
- [x] add tests

Tasks to make it more like existing web component
- [x] block props (see #51 )
- [ ] ~circled props~
- [x] width props (see #51 )
- [ ] ~loading props~
- [x] rename `icon` props to `iconLeft` (see #27 )
- [ ] ~animations on hover~
- [ ] ~animations on click + onBlur state~
- [x] disabled state should decrease opacity and not change it to disabled theme (see #51 )

PR #24, opened 2018-11-16
Button component

RobinCsl:
Relates to #18

### DIFF
--- a/src/Button/ButtonTitle.js
+++ b/src/Button/ButtonTitle.js
@@ -1,0 +1,103 @@
+// @flow
+
+import * as React from 'react';
+import { Text } from 'react-native'; // TODO replace with our own text
+import { defaultTokens } from '@kiwicom/orbit-design-tokens';
+import StyleSheet from '../PlatformStyleSheet';
+
+import type { StylePropType } from '../PlatformStyleSheet/StyleTypes';
+import type { ButtonType } from './ButtonTypes';
+
+type Props = {|
+  +text: React.Node,
+  +type: ButtonType,
+  +style?: StylePropType,
+  +variant?: 'default' | 'sublabel',
+|};
+
+export default function ButtonTitle({
+  text,
+  style,
+  type = 'primary',
+  variant = 'default',
+}: Props) {
+  const variantStyle =
+    variant === 'sublabel' ? styles.sublabel : styles.default;
+  return (
+    <Text style={[styles.common, variantStyle, theme[type].text, style]}>
+      {text}
+    </Text>
+  );
+}
+
+const styles = StyleSheet.create({
+  common: {
+    android: {
+      textTransform: 'uppercase',
+    },
+  },
+  default: {
+    fontWeight: 'bold',
+    fontSize: 16,
+    lineHeight: 21,
+  },
+  sublabel: {
+    fontSize: 14,
+    lineHeight: 18,
+    fontWeight: 'normal',
+  },
+});
+
+const theme = {
+  primary: StyleSheet.create({
+    text: {
+      color: defaultTokens.colorTextButtonPrimary,
+    },
+  }),
+  secondary: StyleSheet.create({
+    text: {
+      color: defaultTokens.colorTextButtonSecondary,
+    },
+  }),
+  info: StyleSheet.create({
+    text: {
+      color: defaultTokens.paletteBlueNormal,
+      web: {
+        color: defaultTokens.colorTextButtonInfo,
+      },
+    },
+  }),
+  success: StyleSheet.create({
+    text: {
+      color: defaultTokens.paletteProductNormal,
+      web: {
+        color: defaultTokens.colorTextButtonSuccess,
+      },
+    },
+  }),
+  warning: StyleSheet.create({
+    text: {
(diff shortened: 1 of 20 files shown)


RobinCsl:
@FilipMessa @tbergq @ferrannp @lukewalczak 
I need to add tests, but I believe the bulk of the Button component is there. Could you have a look and give me feedback? Thanks a lot

src/native/Icon/index.js
```diff
@@ -8,8 +8,8 @@ import iconsMap from './icons.json';
 
 export type Props = {|
   +name: string,
-  +size?: number,
-  +color?: string,
+  size?: number,
```
RobinCsl: I had to do this when cloning an Icon component and flow would rightly complain that I am overriding read-only props otherwise.

src/Button/index.js
```diff
@@ -0,0 +1,117 @@
+// @flow strict
```
tbergq: no strict setup in the project

src/Button/index.js
```diff
@@ -0,0 +1,117 @@
...
+import * as React from 'react';
+import { View } from 'react-native';
+import { defaultTokens } from '@kiwicom/orbit-design-tokens';
+
+import { StyleSheet, type StylePropType } from '..';
+import GenericButtonWrapper from './GenericButtonWrapper';
+import ButtonTitle from './ButtonTitle';
+import type { ButtonType } from './ButtonTypes';
+
+import Icon from '../native/Icon';
```
tbergq: This should be imported like
`import Icon from '@kiwicom/react-native-orbit-icons';`

src/Button/index.js
```diff
@@ -0,0 +1,117 @@
+// @flow strict
+
+import * as React from 'react';
+import { View } from 'react-native';
+import { defaultTokens } from '@kiwicom/orbit-design-tokens';
+
+import { StyleSheet, type StylePropType } from '..';
```
tbergq: I also get a warning about require cycle, so this should 
be imported as

```
import StyleSheet from '../PlatformStyleSheet/index';
import { type StylePropType } from '../PlatformStyleSheet/StyleTypes';

```

src/Button/index.js
```diff
@@ -0,0 +1,117 @@
...
+        </View>
+      </React.Fragment>
+    </GenericButtonWrapper>
+  );
+}
+
+Button.defaultProps = {
+  disabled: false,
+  type: 'primary',
+};
```
tbergq: same as above

src/Button/GenericButtonWrapper.js
```diff
@@ -0,0 +1,118 @@
...
+      accessibilityComponentType="button"
+      accessibilityTraits="button"
+      testID={testID}
+    >
+      {children}
+    </Touchable>
+  );
+}
+
+Button.defaultProps = {
```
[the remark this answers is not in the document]
RobinCsl: Removed default props entirely. Good catch.

src/Button/ButtonTitle.js
```diff
@@ -0,0 +1,99 @@
...
+    android: {
+      textTransform: 'uppercase',
+    },
+  },
+  default: {
+    fontWeight: 'bold',
+    fontSize: 16,
+    lineHeight: 21,
+  },
+  sublabel: { fontSize: 14, lineHeight: 18, fontWeight: 'normal' },
```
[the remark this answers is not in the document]
RobinCsl: You got it.

src/Button/ButtonTitle.js
```diff
@@ -0,0 +1,103 @@
...
+    fontWeight: 'normal',
+  },
+});
+
+const theme = {
+  primary: StyleSheet.create({
+    text: {
+      color: defaultTokens.colorTextButtonPrimary,
+    },
+  }),
```
ferrannp: As there is lot of code, I feel we can save some in several places. I have the feeling you can have just one style for this:

```
const theme = (type) => {
  text: StyleSheet.create({
    color: defaultTokens[`colorTextButton${type.charAt(0).toUpperCase()}${type.slice(1)}`],
  }),
}
```

src/Button/GenericButtonWrapper.js
```diff
@@ -0,0 +1,113 @@
...
+    alignItems: 'center',
+    paddingHorizontal: 8,
+    paddingVertical: 11,
+    height: 44,
+    borderRadius: 6,
+  },
+});
+
+const theme = {
+  primary: StyleSheet.create({
```
ferrannp: I also feel all theses styles can be shorter using a function for `theme` or having an array somehow.

src/Button/ButtonTitle.js
RobinCsl: I would like to use this approach, but sometimes the colors we need to apply for RN do not follow exactly the defaultTokens nomenclature. That's why I had to add `web` for these few exceptions. But I will try to shorten it.

src/Button/GenericButtonWrapper.js
tbergq: Please create a PR to improve @ferrannp 😊

PR #25, opened 2018-11-18
[WIP] :added the Text component

FilipMessa:


src/Text/__tests__/index.test.js
```diff
@@ -2,7 +2,7 @@
 
 import * as React from 'react';
 import { render } from 'react-native-testing-library';
-
+import snapshotDiff from 'snapshot-diff';
```
RobinCsl: What does this do exactly?

src/Text/index.js
```diff
@@ -3,22 +3,65 @@
...
+  text: {
+    margin: 0,
   },
+  italic: {
+    fontStyle: 'italic',
+  },
+  uppercase: {
+    textTransform: 'uppercase',
+  },
+  ...fontWeightGen(),
```
RobinCsl: Maybe with the approach I used, we won't need to generate so many unused styles.

src/Text/__tests__/index.test.js
FilipMessa: TBH I see that Callstack use this so I want to try.
But I should helpful for testing the difference between different component states. 
I followed the pattern that I saw in our codebase. 
I think about like, that if you change something in the rendering based on the status, in the default or extended the component, the test falls and notice you.

src/Text/index.js
FilipMessa: Sure definitely I like your approach so I use your pattern.

PR #30, opened 2018-11-20
Enhance: Button compatible with web-only props

RobinCsl:
I tried to make more platform-specific props; for example: `sublabel` only appears in mobile designs, when on the web, we need to render the component as `a` or `button` tags, depending on whether we have an `href` props.
It is also possible to specify the `width` of the button on both native and web platforms. 
The `block` props is only available on the web: it ignores the width, if it is given, and makes the button full-width; it is full-width on native by default, unless you specify a width.
It seems like I broke the `onClick` handler for the `button` tag on the web. So this must be fixed.


RobinCsl:
I just tried to rewrite history because I did not setup my local git config to have my correct email address... So, sorry that we cannot really see what's going on in this PR.

PR #31, opened 2018-11-21
Add ServiceLogo suffix to TextInput

lukewalczak:


src/TextInput/index.js
```diff
@@ -85,60 +114,70 @@ class TextInput extends React.Component<Props, State> {
...
-                inlineLabel
-                required={required}
-                disabled={disabled}
-              >
-                {label}
-              </FormLabel>
-            </InlineLabel>
+      <TouchableWithoutFeedback onPress={this.focusTextInput}>
+        <View style={styles.inputWrapper}>
+          {label && !inlineLabel && (
```
tbergq: this will crash on android if label = empty string

src/TextInput/index.js
```diff
@@ -16,6 +21,7 @@ type Props = {|
   +inlineLabel?: boolean,
   +label: string,
```
tbergq: Also I think label should accept React.Node, since we most likely will be passing it our Translation component

src/ServiceLogo/component/index.web.js
```diff
@@ -1,18 +1,24 @@
...
     },
   };
+
+  if (typeof size === 'number') {
+    return tokens.height.medium;
+  }
   return tokens.height[size];
 };
+
+const getHeight = height => height;
```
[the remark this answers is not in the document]
lukewalczak: Basically we need that function to set `ServiceLogo` height in web `TextInput`.

https://github.com/kiwicom/universal-orbit-components/blob/475820dc6448c08627b86f21dbdd034f7b7a5fdf/src/TextInput/index.js#L56

https://github.com/kiwicom/universal-orbit-components/blob/475820dc6448c08627b86f21dbdd034f7b7a5fdf/src/ServiceLogo/component/index.web.js#L33

src/TextInput/index.js
lukewalczak: Ok, will check it.

src/TextInput/index.js
```diff
@@ -85,60 +114,70 @@ class TextInput extends React.Component<Props, State> {
...
+                ? styles.inputContainerDisabled
+                : styles.inputContainerDefault,
+              focused
+                ? styles.inputContainerBorderFocused
+                : styles.inputContainerBorderDefault,
             ]}
-          />
+          >
+            {prefix && <Prefix size={size}>{prefix}</Prefix>}
+            {label && inlineLabel && (
```
tbergq: both prefix and label will cause crash on android if they are 0 or empty string, please use prefix != null and label != null(after changing it to React.Node)

src/TextInput/index.js
```diff
@@ -85,60 +114,70 @@ class TextInput extends React.Component<Props, State> {
...
+              placeholderTextColor={defaultTokens.colorPlaceholderInput}
+              editable={!disabled}
+              placeholder={placeholder}
+              value={value}
+              style={[
+                styles.inputField,
+                disabled ? styles.inputFieldDisabled : styles.inputFieldDefault,
+              ]}
+            />
+            {suffix && <Suffix>{suffix}</Suffix>}
```
tbergq: Also potential crash on android

src/TextInput/index.js
lukewalczak: Hmm, I can change it, but for now we don't have `Translation` component for that cases.

src/ServiceLogo/component/index.web.js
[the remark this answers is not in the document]
tbergq: but it just returns the input? What am I missing?

src/TextInput/index.js
tbergq: Yes, the translation component will (probably) never be a part of orbit, but orbit should not be too strictly flow typed, so we should be able to use our translation component

src/ServiceLogo/component/index.web.js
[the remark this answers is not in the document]
lukewalczak: omg, sorry it's a trash

tbergq:
`This branch is out-of-date with the base branch`

Could you please rebase with master?

PR #34, opened 2018-11-22
Support types in text input

lukewalczak:


src/TextInput/index.js
```diff
@@ -79,20 +84,43 @@ class TextInput extends React.Component<Props, State> {
...
+  };
+
+  handleKeyboardType = (type: string) => {
+    if (type === 'number') {
+      return 'numeric';
+    }
+    if (type === 'email') {
+      return 'email-address';
+    }
+    return 'default';
```
ferrannp: Use `if-else` or:

```js
switch(type) {
  case 'number': 
    return 'number'
  case 'email':
    return 'email-address'
  default:
    return 'default'
}
```

src/TextInput/index.js
```diff
@@ -79,20 +84,43 @@ class TextInput extends React.Component<Props, State> {
...
+    onBlur && onBlur();
+  };
+
+  handleChangeText = (value: string) => {
+    const { onChangeText, disabled } = this.props;
+    if (disabled) {
+      return;
+    }
+
+    onChangeText && onChangeText(value);
```
ferrannp: Wow, this is weird, shouldn't it be:

```js
if (!disabled && onChangeText) { 
  onChangeText(value);
}

```

On the other hand, when is disabled (`editable = false`) are you sure this function is not already not called?

src/TextInput/index.js
```diff
@@ -79,20 +84,43 @@ class TextInput extends React.Component<Props, State> {
...
     const { onBlur, disabled } = this.props;
-    if (!disabled) {
-      this.toggleFocus();
-      onBlur && onBlur();
+    if (disabled) {
+      return;
     }
+
+    this.toggleFocus();
+    onBlur && onBlur();
```
ferrannp: Same as `handleChangeText`.

src/TextInput/index.js
```diff
@@ -79,20 +84,43 @@ class TextInput extends React.Component<Props, State> {
...
     const { onFocus, disabled } = this.props;
-    if (!disabled) {
-      this.toggleFocus();
-      onFocus && onFocus();
+    if (disabled) {
+      return;
     }
+
+    this.toggleFocus();
+    onFocus && onFocus();
```
ferrannp: Same as `handleChangeText`. This `onFocus && onFocus();` is just weird, it should be at least:

```js
if (onFocus) {
  onFocus();
}
```

PR #41, opened 2018-11-23
Feat/checkbox component

JoseDuSV:
First version of Checkbox component
- still missing tests and decision about handling onChange event in native version (for now it's just dispatched by 'onPress')

### DIFF
--- a/src/Checkbox/CheckboxIcon.js
+++ b/src/Checkbox/CheckboxIcon.js
@@ -1,0 +1,42 @@
+// @flow
+
+import * as React from 'react';
+import { View } from 'react-native';
+import Icon from '../Icon';
+import StyleSheet from '../PlatformStyleSheet';
+import theme, { parsePxValue } from './styles';
+
+type Props = {|
+  +checked?: boolean,
+  +hasError?: boolean,
+  +disabled?: boolean,
+|};
+
+const styles = StyleSheet.create({
+  box: {
+    alignItems: 'center',
+    justifyContent: 'center',
+    borderWidth: 1,
+    borderStyle: 'solid',
+    borderColor: theme.orbit.borderColorCheckboxRadio,
+    backgroundColor: theme.orbit.backgroundInput,
+    borderRadius: parsePxValue(theme.orbit.borderRadiusNormal),
+    width: parsePxValue(theme.orbit.widthCheckbox),
+    height: parsePxValue(theme.orbit.heightCheckbox),
+  },
+  boxError: {
+    borderColor: theme.orbit.borderColorCheckboxRadioError,
+  },
+});
+
+export default function CheckboxIcon({ checked, hasError, disabled }: Props) {
+  const errorState = hasError && !disabled && !checked;
+  const iconColor = disabled
+    ? theme.orbit.colorIconCheckboxRadioDisabled
+    : theme.orbit.colorIconCheckboxRadio;
+  return (
+    <View style={[styles.box, errorState && styles.boxError]}>
+      {checked && <Icon name="check" color={iconColor} size={14} />}
+    </View>
+  );
+}
--- a/src/Checkbox/CheckboxShared.js
+++ b/src/Checkbox/CheckboxShared.js
@@ -1,0 +1,28 @@
+// @flow
+
+import * as React from 'react';
+import CheckboxText from './CheckboxText';
+import CheckboxIcon from './CheckboxIcon';
+
+type Props = {|
+  +label?: React.Node,
+  +hasError?: boolean,
+  +disabled?: boolean,
+  +checked?: boolean,
+  +info?: React.Node,
+|};
+
+export default function CheckboxShared({
+  label,
+  hasError,
+  disabled,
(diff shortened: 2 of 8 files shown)


src/Checkbox/CheckboxIcon.js
```diff
@@ -0,0 +1,42 @@
...
+    borderRadius: parsePxValue(theme.orbit.borderRadiusNormal),
+    width: parsePxValue(theme.orbit.widthCheckbox),
+    height: parsePxValue(theme.orbit.heightCheckbox),
+  },
+  boxError: {
+    borderColor: theme.orbit.borderColorCheckboxRadioError,
+  },
+});
+
+export default function CheckboxIco({ checked, hasError, disabled }: Props) {
```
RobinCsl:
```suggestion
export default function CheckboxIcon({ checked, hasError, disabled }: Props) {
```

src/Checkbox/CheckboxIcon.js
```diff
@@ -0,0 +1,42 @@
...
+  const errorState = hasError && !disabled && !checked;
+  const iconColor = disabled
+    ? theme.orbit.colorIconCheckboxRadioDisabled
+    : theme.orbit.colorIconCheckboxRadio;
+  return (
+    <View style={[styles.box, errorState && styles.boxError]}>
+      {checked && <Icon name="check" color={iconColor} size={14} />}
```
RobinCsl: I am not sure we should leave the size being given like this. Does 14 match the size of the font as well? If that's the case, we could take it from `defaultTokens`.

src/Checkbox/CheckboxShared.js
```diff
@@ -0,0 +1,28 @@
+// @flow
+
+import * as React from 'react';
+import CheckboxText from './CheckboxText';
+import CheckboxIco from './CheckboxIcon';
```
RobinCsl:
```suggestion
import CheckboxIcon from './CheckboxIcon';
```

src/Checkbox/CheckboxShared.js
```diff
@@ -0,0 +1,28 @@
...
+export default function CheckboxShared({
+  label,
+  hasError,
+  disabled,
+  checked,
+  info,
+}: Props) {
+  return (
+    <>
+      <CheckboxIco checked={checked} hasError={hasError} disabled={disabled} />
```
RobinCsl:
```suggestion
      <CheckboxIcon checked={checked} hasError={hasError} disabled={disabled} />
```

src/Checkbox/CheckboxText.js
```diff
@@ -0,0 +1,48 @@
...
+
+const styles = StyleSheet.create({
+  view: {
+    display: 'flex',
+    flex: 1,
+    flexDirection: 'column',
+    marginStart: parsePxValue(theme.orbit.spaceXSmall),
+  },
+  label: {
+    fontFamily: Platform.OS === 'web' ? theme.orbit.fontFamily : 'Roboto',
```
RobinCsl: I believe this is a very general "problem" to solve, so we should maybe extract it into a helper function?

src/Checkbox/styles/index.js
```diff
@@ -0,0 +1,12 @@
+// @flow
+
+import { defaultTokens } from '@kiwicom/orbit-design-tokens';
+
+export const parsePxValue = (pxString: string) => parseFloat(pxString);
```
RobinCsl: I like that 🎉

src/Checkbox/CheckboxText.js
RobinCsl: You're right, so no need to refactor this now.

src/Checkbox/CheckboxIcon.js
JoseDuSV: Good point. Currently the size matches with font, but I don't think it will be rule all the time (because for example size of box is set by separate values).
In current `orbit-components` there is svg check icon with fixed 16px width and height. So it's currently also not really dynamic.
We can definitely consider to make it more responsive/dynamic, but it will probably require to extend `defaultTokens`.

src/Checkbox/CheckboxShared.js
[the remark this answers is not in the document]
JoseDuSV: Yep, it's already fixed by Robin.

src/Checkbox/styles/index.js
```diff
@@ -0,0 +1,12 @@
...
+export const parseStringToFloat = (stringValue: string) =>
+  parseFloat(stringValue);
+
```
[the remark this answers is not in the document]
JoseDuSV: Basically I currently treat all style related stuff more like 'work in progress' which will be probably replaced by some global solution. 
Thats why I simply wanted to have all style parsing related stuff in one place. So it will be easier to update/replace it in a future.
And I'm also still not really sure about using just `parseFloat` for design-tokens values parsing, so thats another reason why I wanted to have it wrapped, so it can be easily switched when we decide to use some different approach.

PR #51, opened 2018-11-27
Enhance: Button compatible with web-only props

RobinCsl:
I tried to make more platform-specific props; for example: sublabel only appears in mobile designs, when on the web, we need to render the component as a or button tags, depending on whether we have an href props.
It is also possible to specify the width of the button on both native and web platforms.
The block props is only available on the web: it ignores the width, if it is given, and makes the button full-width; it is full-width on native by default, unless you specify a width.


src/Button/Button.stories.js
```diff
@@ -42,14 +51,24 @@ storiesOf('Button', module)
       ['undefined', ...Object.keys(icons)],
       'calendar'
     );
+    const width = number('Width', NaN);
```
lukewalczak: I think it's better to set default value instead of `Nan`. Check if you can change `Nan` value in mobile playground.

RobinCsl: My thinking was that you can see it defaults to being full-width when you do not specify any width. But I guess it's weird to see `NaN` as a default value 👍

RobinCsl:
```suggestion
    const width = number('Width', 300);
```

PR #56, opened 2018-11-27
Add clear button in text input

lukewalczak:


src/TextInput/index.js
```diff
@@ -185,8 +205,16 @@ class TextInput extends React.Component<Props, State> {
...
-
-            {suffix != null && <Suffix>{suffix}</Suffix>}
+            {Platform.OS !== 'web' && (
+              <ClearButton
+                onPress={this.clearValue}
+                focused={focused}
+                value={value}
+              />
+            )}
+            {suffix != null && Platform.OS === 'web' && (
```
RobinCsl: So that means I cannot see the service logo anymore. I think, even though it's not in the design, we had a use case for a payment form. But now that I think of it, it was actually prefixing the input field, and not suffixing. So maybe we can leave it like this.
@tbergq thoughts?

src/TextInput/index.js
```diff
@@ -50,6 +54,17 @@ const Suffix = ({ children }) => {
   return <View style={styles.suffix}>{suffix}</View>;
 };
 
+const ClearButton = ({ onPress, focused, value }) => {
+  if (focused && value) {
```
RobinCsl: I know this is what the design says, but I find it weird that I cannot reset the input field if it's not in focused state and there is a value in it. I need two touches to reset the input field. WDYT?
@tbergq thoughts?

src/TextInput/index.js
lukewalczak: >  it was actually prefixing the input field, and not suffixing.

I had the same feelings, that's why removed it from mobile.

src/TextInput/index.js
lukewalczak: In the majority of the inputs, clear button appears when input is focused and value.length > 0.

tbergq: I would say stick to design, and rather create an issue with designers

PR #60, opened 2018-11-27
Ignore require cycle from node-modules

lukewalczak:


index.js
```diff
@@ -1,5 +1,8 @@
 // @flow
 
+import { YellowBox } from 'react-native';
 import StorybookUI from './storybook.native';
 
+YellowBox.ignoreWarnings(['Require cycle: node_modules']);
```
RobinCsl:
```suggestion
YellowBox.ignoreWarnings(['Require cycle: node_modules/@storybook']);
```
I would prefer to only ignore the ones coming from `@storybook` since this is the only cause for these warnings at the moment.

PR #62, opened 2018-11-27
Add: Stepper Component

### LINKED ISSUE #55: Create Stepper component
Cf Figma

This should amount to extracting 

https://github.com/kiwicom/mobile/blob/master/packages/shared/src/buttons/IncrementDecrementButtons.js

RobinCsl:
Closes #55


src/Stepper/index.js
```diff
@@ -0,0 +1,74 @@
...
+  onDecrement,
+  number,
+  showNumber = true,
+  numberStyle,
+  min,
+  max,
+  disableDecrement: originalDisableDecrement,
+  disableIncrement: originalDisableIncrement,
+}: Props) {
+  const disableDecrement = (min && number <= min) || originalDisableDecrement;
```
lukewalczak: `min && number <= min` and `max && number >= max` are used in some places. Maybe it's worth to extract it to const?

RobinCsl: Good idea, I'll do that!

src/Stepper/index.js
```diff
@@ -0,0 +1,76 @@
...
+  onDecrement,
+  number,
+  showNumber = true,
+  numberStyle,
+  min,
+  max,
+  disableDecrement: originalDisableDecrement,
+  disableIncrement: originalDisableIncrement,
+}: Props) {
+  const belowMin = min && number <= min;
```
tbergq: Would not `number <= min` be sufficient since
1 <= undefined -> `false`

src/Stepper/index.js
```diff
@@ -0,0 +1,76 @@
...
+  const aboveMax = max && number >= max;
+  const disableDecrement = belowMin || originalDisableDecrement;
+  const disableIncrement = aboveMax || originalDisableIncrement;
```
tbergq: `||` -> `??`

src/Stepper/index.js
RobinCsl: I would get the following Flow error:
```
Cannot compare number [1] to undefined [2]. (References: [1] [2])
```

src/Stepper/StepperButton.js
```diff
@@ -0,0 +1,49 @@
...
+import { defaultTokens } from '@kiwicom/orbit-design-tokens';
+
+import Icon from '../Icon';
+import Touchable from '../Button/Touchable';
+import StyleSheet from '../PlatformStyleSheet';
+import type { StylePropType } from '../PlatformStyleSheet/StyleTypes';
+
+type ButtonProps = {|
+  +icon: React.Element<typeof Icon>,
+  +touchable: boolean,
```
ferrannp: Does `touchable` need to be mandatory prop? Can we have it optional with default value?

src/Stepper/index.js
```diff
@@ -0,0 +1,76 @@
...
+  max,
+  disableDecrement: originalDisableDecrement,
+  disableIncrement: originalDisableIncrement,
+}: Props) {
+  const belowMin = min && number <= min;
+  const aboveMax = max && number >= max;
+  const disableDecrement = belowMin ?? originalDisableDecrement;
+  const disableIncrement = aboveMax ?? originalDisableIncrement;
+
+  const commonIconProps = {
```
ferrannp: Why common? It's not exported no?

src/Stepper/StepperButton.js
RobinCsl: Good point.

src/Stepper/index.js
RobinCsl: Renamed `commonIconProps` to `iconProps`. Thanks

PR #79, opened 2018-11-29
Loader and Page Loader components

lukewalczak:
- [x] - add test
- [x] - export components in main index


src/Loader/__tests__/index.test.js
```diff
@@ -0,0 +1,33 @@
...
+describe('Loader', () => {
+  it('should match snapshot diff between small and large loader', () => {
+    const small = render(<Loader size="small" />);
+    const large = render(<Loader size="large" />);
+
+    expect(snapshotDiff(small, large, { contextLines: 1 })).toMatchSnapshot();
+  });
+});
+
+describe('Page Loader', () => {
```
RobinCsl: I think the following tests have a mistake in them, since the same conditional is used. Besides, this means that will be dependent on the platform the tests are run on. Would it make sense to impose `Platform.OS = 'android'` for the first test, and `Platform.OS = 'ios'` for the second?

lukewalczak: sure, will adjust it!

PR #81, opened 2018-11-30
Enhance: Checkbox special states styles + tests

JoseDuSV:
Added tests and mainly web hover (+ active, focus) state handling. If we decide to use it in more cases, `Hoverable` can be probably extracted as separate component. 
I'm also not that happy with the current `HoverMonitor` solution, which can be probably expanded, but currently it serves its purpose.


src/Checkbox/Checkbox.web.js
```diff
@@ -33,16 +20,15 @@ export default function Checkbox({
   name,
   info,
   onChange,
-}: Props) {
-  const labelStyle = {
+}: CheckboxProps) {
+  const styleLabel = {
```
RobinCsl:
```suggestion
  const labelStyle = {
```
I think we prefer this way of writing variables, but I am curious to know your reason for that change. Do you think it could be problematic later on?

src/Checkbox/Checkbox.web.js
```diff
@@ -33,16 +20,15 @@ export default function Checkbox({
...
     display: 'flex',
     flexDirection: 'row',
     cursor: disabled ? 'default' : 'pointer',
-    opacity: disabled ? theme.orbit.opacityCheckboxDisabled : 1,
   };
 
   return (
-    <label style={labelStyle}>
+    <label style={styleLabel}>
```
RobinCsl:
```suggestion
    <label style={labelStyle}>
```

src/Checkbox/CheckboxShared.js
```diff
@@ -1,28 +1,98 @@
...
-  );
+class CheckboxShared extends React.Component<
+  CheckboxSharedProps,
+  CheckboxSharedState
+> {
+  constructor(props: CheckboxSharedProps) {
+    super(props);
+
+    this.state = {
+      showFocus: false,
```
RobinCsl:
```suggestion
      focused: false,
```
maybe to be a bit more consistent

src/Checkbox/CheckboxShared.js
```diff
@@ -1,28 +1,98 @@
...
+  handleOnMouseLeave = () => {
+    this.setState({ hovered: false });
+  };
+
+  handleOnPressIn = () => {
+    this.setState({ pressed: true });
+  };
+
+  handleOnPressOut = () => {
+    this.setState({ pressed: false, showFocus: false });
```
RobinCsl:
```suggestion
    this.setState({ pressed: false, focused: false });
```

src/Checkbox/CheckboxShared.js
```diff
@@ -1,28 +1,98 @@
...
+  };
+
+  handleOnFocus = () => {
+    this.setState({ showFocus: true });
```
RobinCsl:
```suggestion
    this.setState({ focused: true });
```

src/Checkbox/CheckboxShared.js
```diff
@@ -1,28 +1,98 @@
...
+  };
+
+  handleOnBlur = () => {
+    this.setState({ showFocus: false });
```
RobinCsl:
```suggestion
    this.setState({ focused: false });
```

src/Checkbox/CheckboxShared.js
```diff
@@ -1,28 +1,98 @@
...
+  };
+
+  render() {
+    const { label, hasError, disabled, checked, info, onPress } = this.props;
+    const { showFocus, hovered, pressed } = this.state;
```
RobinCsl:
```suggestion
    const { focused, hovered, pressed } = this.state;
```

src/Checkbox/CheckboxShared.js
```diff
@@ -1,28 +1,98 @@
...
+          onFocus={this.handleOnFocus}
+          onBlur={this.handleOnBlur}
+          disabled={disabled}
+        >
+          <View style={[styles.view, disabled && styles.viewDisabled]}>
+            <CheckboxIcon
+              checked={checked}
+              hasError={hasError}
+              disabled={disabled}
+              focused={showFocus}
```
RobinCsl:
```suggestion
              focused={focused}
```

src/Checkbox/HoverMonitor.js
```diff
@@ -0,0 +1,32 @@
...
+  }
+
+  function disableHover() {
+    if (hoverEnabled) {
+      hoverEnabled = false;
+    }
+  }
+
+  if (typeof document !== 'undefined') {
+    document.addEventListener('touchstart', disableHover, true);
```
RobinCsl: So this disables your abstraction of `hover` for the whole document, thanks to your last argument being `true`, right?
I think it is quite clever. 
My only critic would be that we might need to clean up these listeners after the component where it is used is unmounted. What do you think?

src/Checkbox/Checkbox.web.js
JoseDuSV: No, you are right. It makes more sense this way. I probably just used this pattern because of switching from `react-native` styling which is usually something like `styles.sublabel`.

src/Checkbox/HoverMonitor.js
JoseDuSV: Right.
Cleaning listeners when there is no `Hoverable` component in use is one of the possible upgrades. Also mouse listener can be probably added after first `touchstart`. Because it's not needed before that.

src/Checkbox/CheckboxShared.js
JoseDuSV: I've used `focused` name at first. But I've changed it to prevent confusion with actual focus state of component. Since it serves only for switching visual style of checkbox focus border.
Which is turned off by clicking on component, event that the component is still in focus (`focused`).
Hope it makes sense, basically it just prevents checkbox being stuck with thick blue border when focused-on with keyboard and then clicked.

RobinCsl: I see. So maybe the name should be changed to `focusDisplayed` or something similar. What do you think?

src/Checkbox/HoverMonitor.js
RobinCsl: Oh right, when there is no _other_ `Hoverable` simultaneously... So it's a bit more involved than what I originally thought. So we will do these improvements later. 
Could you still extract these two possible improvements into issues? Thanks

src/Checkbox/CheckboxShared.js
JoseDuSV: Definitely, `focusDisplayed` sounds good. Lot better than original function-like `showFocus`.

Updated

src/Checkbox/HoverMonitor.js
JoseDuSV: Ok, issue for `Hoverable` enhancements created.

PR #85, opened 2018-11-30
Add: Rating component

### LINKED ISSUE #77: Create Rating component
Extract https://github.com/kiwicom/mobile/blob/master/packages/shared/src/Stars.js

JoseDuSV:
Closes #77

### DIFF
--- a/src/Rating/Rating.js
+++ b/src/Rating/Rating.js
@@ -1,0 +1,21 @@
+// @flow
+
+import * as React from 'react';
+import Text from '../Text';
+import StyleSheet from '../PlatformStyleSheet';
+import type { StylePropType } from '../PlatformStyleSheet/StyleTypes';
+
+type Props = {|
+  +rating?: number,
+  +style?: StylePropType,
+|};
+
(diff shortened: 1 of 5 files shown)


src/Rating/Rating.js
```diff
@@ -0,0 +1,21 @@
...
+import Text from '../Text';
+import StyleSheet from '../PlatformStyleSheet';
+import type { StylePropType } from '../PlatformStyleSheet/StyleTypes';
+
+type Props = {|
+  +rating?: ?number,
+  +style?: StylePropType,
+|};
+
+export default function Rating({ rating, style }: Props) {
```
RobinCsl: I think I would prefer to have 
```suggestion
export default function Rating({ rating = 0, style }: Props) {
```
and not have `|| 0` below, for the logic to be a bit cleaner. And the type can now be
`+rating?: number`.
What do you think?

JoseDuSV: Right, I'm also using default values in other components so it's definitely cleaner this way.

PR #94, opened 2018-12-05
Add: CarrierLogo component

### LINKED ISSUE #88: Add CarrierLogo component
This is on `orbit-components`:
https://kiwicom.github.io/orbit-components/?selectedKind=CarrierLogo&selectedStory=One%20carrier&full=0&addons=1&stories=1&panelRight=0&addonPanel=storybook%2Factions%2Factions-panel

In `mobile` repository, this was done like this
https://github.com/kiwicom/mobile/blob/master/app/MMB/src/scenes/tripOverview/TimelineDeparture.js#L108-L109
but I think the `orbit-components` way is probably better?

JoseDuSV:
Closes #88

### DIFF
--- a/src/CarrierLogo/CarrierLogo.js
+++ b/src/CarrierLogo/CarrierLogo.js
@@ -1,0 +1,77 @@
+// @flow
+
+import * as React from 'react';
+import { View, Image, PixelRatio } from 'react-native';
+import { defaultTokens } from '@kiwicom/orbit-design-tokens';
+import StyleSheet from '../PlatformStyleSheet';
+import type { Props } from './CarrierLogoTypes';
+import { SIZE_OPTIONS, CARRIER_TYPE_OPTIONS, BASE_URL } from './consts';
+import { logoSizes } from './styles';
+
+export default function CarrierLogo({
+  size = SIZE_OPTIONS.LARGE,
+  carriers = [],
+}: Props) {
+  const carriersLength = carriers.length;
+  const urlSize =
+    (carriersLength > 1 || size === SIZE_OPTIONS.SMALL ? 16 : 32) *
+    (PixelRatio.get() >= 2 ? 2 : 1);
+
+  return (
+    <View
+      style={[
+        styles.wrapper,
+        carriersLength > 1 ? sizeStyles[SIZE_OPTIONS.LARGE] : sizeStyles[size],
+      ]}
+    >
+      {carriers.slice(0, 4).map((carrierImage, index) => (
+        <Image
+          key={carrierImage.code}
+          style={[
+            styles.logo,
+            getLogoSizeStyle(carriersLength, size),
+            carriersLength === 2 && index === 1 && styles.logoFlexEnd,
+          ]}
+          source={{
+            uri: `${BASE_URL}/airlines/${urlSize}/${
+              carrierImage.code
+            }.png?default=${
+              carrierImage.type === undefined
+                ? CARRIER_TYPE_OPTIONS.AIRLINE
+                : carrierImage.type
+            }.png`,
+          }}
+        />
+      ))}
+    </View>
+  );
+}
+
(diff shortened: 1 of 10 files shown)


src/CarrierLogo/CarrierLogo.js
```diff
@@ -0,0 +1,77 @@
...
+      {carriers.slice(0, 4).map((carrierImage, index) => (
+        <Image
+          key={carrierImage.code}
+          style={[
+            styles.logo,
+            getLogoSizeStyle(carriersLength, size),
+            carriersLength === 2 && index === 1 && styles.logoFlexEnd,
+          ]}
+          source={{
+            uri: `${BASE_URL}/airlines/${urlSize}/${
```
RobinCsl: Could you extract this URL generation into a function and write a test for it too?

src/CarrierLogo/CarrierLogo.js
```diff
@@ -0,0 +1,77 @@
...
+  return (
+    <View
+      style={[
+        styles.wrapper,
+        carriersLength > 1 ? sizeStyles[SIZE_OPTIONS.LARGE] : sizeStyles[size],
+      ]}
+    >
+      {carriers.slice(0, 4).map((carrierImage, index) => (
+        <Image
+          key={carrierImage.code}
```
RobinCsl: This might be a problem if duplicate carrier is passed in `carriers` prop.

src/CarrierLogo/CarrierLogo.js
```diff
@@ -0,0 +1,77 @@
...
+import { logoSizes } from './styles';
+
+export default function CarrierLogo({
+  size = SIZE_OPTIONS.LARGE,
+  carriers = [],
+}: Props) {
+  const carriersLength = carriers.length;
+  const urlSize =
+    (carriersLength > 1 || size === SIZE_OPTIONS.SMALL ? 16 : 32) *
+    (PixelRatio.get() >= 2 ? 2 : 1);
```
RobinCsl: Could you explain why the PixelRatio will always be maximum 2?

src/CarrierLogo/CarrierLogo.js
RobinCsl: Also, maybe in a future iteration, we would want these assets to be available as assets for mobile apps, when on the web they could be fetched from URL. WDYT?

src/CarrierLogo/CarrierLogo.js
JoseDuSV: I've been thinking about this. But since it's used this way in `orbit` https://github.com/kiwicom/orbit-components/blob/master/src/CarrierLogo/index.js , I've assumed that it cannot happen.
Do you think there will be case when we want to display same logo multiple times? In this case we need to use something else as a key.
Otherwise, I can at least add some data parsing to remove duplicate carriers (if it's desired behaviour).

src/CarrierLogo/CarrierLogo.js
JoseDuSV: Sure, I'll extract it. 
And we can definitely extend it later and bundle the assets. I'm just not sure how often assets like carrier logos changes and expands.

src/CarrierLogo/CarrierLogo.js
JoseDuSV: This way I'm handling just retina/non-retina displays.
But I see that there is even bigger version of assets available, so I'll edit and extend it even for `@3x` cases.

src/CarrierLogo/CarrierLogo.js
JoseDuSV: Helper extracted and tests added.

src/CarrierLogo/CarrierLogo.js
JoseDuSV: Just to be sure I've added parser which filters out carriers with the same `code` (which serves as id). Also separated it into helpers and added some test for it.

PR #98, opened 2018-12-05
Add: StopoverArrow component

### LINKED ISSUE #93: Create the StopoverArrow component
# StopoverArrow

### props
- stops: *number* (default value 0, max 4)

#### Without stopover
screen shot 2018-12-04 at 12 57 37
#### With stopover
screen shot 2018-12-04 at 12 57 31

jvaclavik:
closes #93

### DIFF
--- a/android/app/build.gradle
+++ b/android/app/build.gradle
@@ -140,6 +140,7 @@ dependencies {
     implementation fileTree(dir: "libs", include: ["*.jar"])
     implementation "com.android.support:appcompat-v7:${rootProject.ext.supportLibVersion}"
     implementation "com.facebook.react:react-native:+"  // From node_modules
+    implementation project(':react-native-svg')
 }
 
 // Run this once to be able to run the application with BUCK
--- a/android/app/src/main/java/com/storybooknative/MainApplication.java
+++ b/android/app/src/main/java/com/storybooknative/MainApplication.java
@@ -8,6 +8,8 @@ import com.facebook.react.ReactPackage;
 import com.facebook.react.shell.MainReactPackage;
 import com.facebook.soloader.SoLoader;
 
+import com.horcrux.svg.SvgPackage;
+
 import java.util.Arrays;
 import java.util.List;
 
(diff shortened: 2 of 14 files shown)


android/app/build.gradle
```diff
@@ -140,6 +140,7 @@ dependencies {
     implementation fileTree(dir: "libs", include: ["*.jar"])
     implementation "com.android.support:appcompat-v7:${rootProject.ext.supportLibVersion}"
     implementation "com.facebook.react:react-native:+"  // From node_modules
+    implementation project(':react-native-svg')
```
RobinCsl: I believe we wanted to stay away from `react-native-svg` as it drains batteries. @tbergq could you confirm?

RobinCsl: I think it kind of is, but there were other requirements for this arrow: the marks on the arrow could be more numerous, and they could be square.

JoseDuSV: And what about building arrow just from Views for line and stop squares, with `chevron-right` icon on right side? To stay away from native dependencies.

screenshot 2018-12-07 at 09 54 56

tbergq: How about requesting this source as a png file, our see if there is some online tool converting svg -> png?

RobinCsl: We could do this on Figma

JoseDuSV: @jvaclavik Right, it will be quite a lot of objects when used in list.
As I see in designs the arrow probably doesn't need to be responsive and have a fixed width.
So @tbergq solution using just images seems to me as simplest one.

jvaclavik:
We can probably decline this PR and close the issue (#93), because this component will not be in new design.

RobinCsl:
@FilipMessa What's the status of the design with this StopOverArrow?

FilipMessa:
@RobinCsl 
There will be new designs without stopover arrow, so we can close this.
You can find it here: https://www.figma.com/file/BiZpzn612rlJ85vAozQkRtwC/Results-Listing?node-id=3694%3A30885

PR #101, opened 2018-12-07
Add: Icons available on Web (temporary)

RobinCsl:
This is a bit hacky, but we can now see icons in the storybook for web.
To use the icon of this component library inside a web project, we would need to install `@kiwicom/orbit-components` and `styled-components` as dependencies I believe. This would need to be tested and made precise.

### DIFF
--- a/src/Icon/Icon.native.js
+++ b/src/Icon/Icon.native.js
@@ -2,15 +2,11 @@
 
 import * as React from 'react';
 import { Text as RNText } from 'react-native';
+import { defaultTokens } from '@kiwicom/orbit-design-tokens';
 import StyleSheet from '../PlatformStyleSheet';
 
 import iconsMap from './icons.json';
-
-export type Props = {|
-  +name: string,
-  size?: number,
-  color?: string,
-|};
+import type { Props } from './IconTypes';
 
 const getIconCharacter = name => {
   const icon = iconsMap[name];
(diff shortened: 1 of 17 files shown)


src/Icon/Icon.native.js
```diff
@@ -50,3 +48,12 @@ const styles = StyleSheet.create({
     },
   },
 });
+
+const iconSize = (size: Size) =>
+  StyleSheet.create({
```
JoseDuSV: It would be nice to avoid recreating stylesheet with every render.

What about adding something like below to `styles`
```
small: {
    fontSize: parseFloat(defaultTokens.widthIconSmall),
},
medium: {
    fontSize: parseFloat(defaultTokens.widthIconMedium),
},
large: {
    fontSize: parseFloat(defaultTokens.widthIconLarge),
},
```
and then use it like
`<RNText style={[styles.icon, styles[size], { color }]}>`

I understand that this approach can seems to look more 'hardcoded', but I don't think we will be expanding size options anytime soon.

JoseDuSV:
`TextInput` with icon currently crashing in RN, changing

https://github.com/kiwicom/universal-components/blob/42ee1080a254a54cb0d3953a2c2a09b2495c0bb2/src/TextInput/TextInput.js#L39-L42

to below should solve it

```
return React.cloneElement(child, {
   color: iconColor,
   size,
});
```

RobinCsl:
This is only temporary; see #112

PR #102, opened 2018-12-07
Notification component

lukewalczak:
alert


package.json
```diff
@@ -32,7 +32,8 @@
   "author": "kiwi.com",
   "license": "MIT",
   "dependencies": {
-    "@kiwicom/orbit-design-tokens": "^0.2.5"
+    "@kiwicom/orbit-design-tokens": "^0.2.5",
+    "react-native-status-bar-height": "^2.2.0"
```
ferrannp: Do we really need this library + the block in webpack.config.js ?

src/Alert/Alert.js
```diff
@@ -0,0 +1,235 @@
...
+  handleLayout = (e: OnLayout) => {
+    const { layout } = this.state;
+    const { height } = e.nativeEvent.layout;
+    const { measured } = layout;
+    const { isOpen, translateY, opacity } = this.state;
+
+    this.setState({ layout: { height, measured: true } }, () => {
+      if (measured) {
+        if (!isOpen) {
+          // - 300 is mainly for displaying in storybook
```
ferrannp: Not fully understand this but if it's related to Storybook... It probably should not be an extra case.

src/Alert/Alert.js
```diff
@@ -0,0 +1,235 @@
...
+
+  _toggleAlert(type: NotificationType, message: React.Node | string) {
+    const { isOpen } = this.state;
+    if (!isOpen) {
+      this.setState({
+        isOpen: true,
+        alertType: type,
+        alertMessage: message,
+      });
+      this.showAlert();
```
ferrannp: Maybe call this as part of the setState callback?

src/Warning/Warning.js
```diff
@@ -0,0 +1,271 @@
...
+    height: number,
+    measured: boolean,
+  },
+  opacity: Animated.Value,
+  translateY: Animated.Value,
+  isOpen: boolean,
+  warningType: NotificationType,
+  warningTitle: React.Node | string,
+  warningMessage: React.Node | string,
+|};
```
ferrannp: I think you can unify types with Alerts. Have some `type AlertBase` and then do `AlertBase & WarningStateType`.

src/Warning/Warning.js
```diff
@@ -0,0 +1,271 @@
...
+
+const ANIMATION_DURATION = 250;
+
+const AnimatedTouchable = Animated.createAnimatedComponent(Touchable);
+
+export default class Warning extends React.Component<Props, State> {
```
ferrannp: This component seems the same or 90% the same as `Alert.js`. I think abstraction is required.

package.json
lukewalczak: - changes in webpack.config are needed for our storybook web which cannot compile without it.

- without that lib I will have to wrap component with `SafeArea` for `ios` and get `currentHeight` for `android`, that's why I think better option is to use it.

src/Alert/Alert.js
lukewalczak: Only the value `-300` is needed for proper displaying in storybook, but it doesnt affect the component behaviour.

src/Warning/Warning.js
lukewalczak: Will do that.

src/Warning/Warning.js
lukewalczak: Will try to extract common codebase or create just one component and distinguish `Alert` and `Warning` using props.

src/Alert/Alert.js
lukewalczak: Rendering `null` when `isOpen === false`

src/Notification/Warning.js
```diff
@@ -0,0 +1,117 @@
...
+  +style?: StylePropType,
+  +notificationStyle: NotificationStyleType,
+  +warningTitle: React.Node | string,
+  +warningMessage: React.Node | string,
+  +onPress: () => void,
+|};
+
+const AnimatedTouchable = Animated.createAnimatedComponent(Touchable);
+
+export default function Warning({
```
ferrannp: Question, why this component is `Warning` but it can be `error` or `success` ? Isn't that a bit confusing ? Or web one in `orbit-components` is the same?

src/Notification/Alert.js
```diff
@@ -0,0 +1,36 @@
...
+}
+
+const styles = StyleSheet.create({
+  container: {
+    justifyContent: 'center',
+    alignItems: 'center',
+  },
+  content: {
+    fontSize: 14,
+    color: 'white',
```
RobinCsl: Can this come from defaultTokens?

src/Notification/Alert.js
```diff
@@ -0,0 +1,36 @@
...
+    paddingVertical: 8,
+    paddingHorizontal: 8,
```
RobinCsl: `padding: 8` would be enough?

src/Notification/Notification.js
```diff
@@ -0,0 +1,287 @@
...
+    },
+    opacity: new Animated.Value(0),
+    translateY: new Animated.Value(0),
+    isOpen: false,
+    notificationStyle: 'error',
+    notificationTitle: '',
+    notificationMessage: '',
+  };
+
+  _lastCallTimestamp = null;
```
RobinCsl: Is there a specific reason to use underscores? I think we should add this as an eslint rule to prevent them. What do you think?

src/Notification/Warning.js
lukewalczak: Currently you have to specify what type of `Notification` (`alert` or `warning`) you would like to use ( it's based on `figma` design). Then you have to specify notification style (`error`, `warning` or `success`.

src/Notification/Notification.js
lukewalczak: In that file I have two `toggleAlert` functions where one is internal (with `underscore) and second is external. I just add these `underscores` to keep the convention. I'm also not a fan of using underscores, but in that case it was the best idea imo.

src/Notification/Notification.js
```diff
@@ -0,0 +1,287 @@
...
+        toValue: 1,
+        duration: ANIMATION_DURATION,
+        useNativeDriver: true,
+      }),
+      Animated.timing(translateY, {
+        toValue: 0,
+        duration: ANIMATION_DURATION,
+        useNativeDriver: true,
+      }),
+    ]).start(() => {});
```
RobinCsl: Is this empty callback necessary?

lukewalczak: Nope

src/Notification/Warning.js
RobinCsl: I could not find any similar component on `orbit-components`. 
I can see why the name could be confusing. 
Maybe we could have ImportantNotification and InformativeNotification. I don't know. What do you think?

src/Notification/Warning.js
```diff
@@ -0,0 +1,117 @@
...
+          <Icon name={iconRight} color={iconColor} size={iconSize} />
+        </View>
+      </View>
+    </AnimatedTouchable>
+  );
+}
+
+const styles = StyleSheet.create({
+  header: {
+    fontWeight: '700',
```
RobinCsl: There might be some values in `defaultTokens` that you might use.

storybook.native/story-loader.js
```diff
@@ -0,0 +1,45 @@
+// @flow
```
RobinCsl: Somehow, this should have been gitignored. Can you inquire why this was not the case? Thank you!

src/Notification/Notification.js
RobinCsl: Maybe a way to make it very explicit would be to replace the underscore with `internal`. I don't have a strong opinion about it.

src/Notification/ImportantNotification.js
```diff
@@ -0,0 +1,117 @@
...
+  +warningMessage: React.Node | string,
+  +onPress: () => void,
+|};
+
+const AnimatedTouchable = Animated.createAnimatedComponent(Touchable);
+
+export default function ImportantNotification({
+  style,
+  onPress,
+  warningTitle,
```
RobinCsl: I believe these `warning` prefixes are now artefacts of the preview naming convention. Could you update them, pretty please?

PR #104, opened 2018-12-10
Add Tooltip

### LINKED ISSUE #50: Create the Tooltip component.
screen shot 2018-11-26 at 15 34 16

jvaclavik:
Closes #50.

There are two components:
- `Tooltip`: Calculates position of children element on the screen and handles how and where should be rendered `TooltipBubble` for this element
- `TooltipBubble`: Component which deals with styling of the bubble and its arrow according to props


src/Tooltip/Tooltip.js
```diff
@@ -0,0 +1,106 @@
...
+      this.childrenRef.measure((a, b, width, height) => {
+        const roundedHeight = Math.round(height);
+        const { childrenHeight } = this.state;
+        roundedHeight !== childrenHeight &&
+          this.setState({ childrenHeight: roundedHeight });
+      });
+  };
+
+  onLayout = (event: ViewLayoutEvent) => {
+    const { layout } = event.nativeEvent;
```
RobinCsl: I don't know if this is necessary, but it could be worth it to setState only if the value is different. What do you think?

src/Tooltip/Tooltip.js
```diff
@@ -0,0 +1,106 @@
...
+    const { elementY, childrenHeight } = this.state;
+    const isArrowDown = screenHeight / 2 > elementY;
+    const computedStyles = getComputedStyles(isArrowDown);
+
+    return (
+      <View onLayout={this.onLayout} style={styles.tooltipComponent}>
+        <View ref={this.getChildrenRef} renderToHardwareTextureAndroid>
+          {children}
+        </View>
+        {isActive && childrenHeight !== 0 && (
```
RobinCsl: Could you give this condition a name? Maybe something like
```
const displayTooltipBubble = isActive && childrenHeight !== 0;
```

src/Tooltip/Tooltip.js
```diff
@@ -0,0 +1,106 @@
...
+          <TooltipBubble
+            text={text}
+            arrowVerticalPosition={isArrowDown ? 'flex-start' : 'flex-end'}
```
RobinCsl: Can this logic be extracted before the component's render method?

src/Tooltip/Tooltip.js
```diff
@@ -0,0 +1,106 @@
...
+            arrowHorizontalPosition={arrowHorizontalPosition}
+            style={computedStyles.tooltipBubble}
+            verticalShift={verticalShift + childrenHeight}
```
RobinCsl: Likewise here.

src/Tooltip/Tooltip.js
```diff
@@ -0,0 +1,106 @@
...
+}
+
+const styles = StyleSheet.create({
+  tooltipComponent: {
+    position: 'relative',
+    zIndex: 100,
+  },
+});
+
+const getComputedStyles = isArrowDown =>
```
RobinCsl: @tbergq I am still not sure what the best way is to avoid recreating styles on every render? Is that enough because `StyleSheet.create` will somehow cache the styles?

src/Tooltip/TooltipBubble.js
```diff
@@ -0,0 +1,89 @@
...
+  arrow: {
+    backgroundColor: defaultTokens.backgroundBadgeDark,
+    borderRadius: parseFloat(defaultTokens.borderRadiusSmall),
+    height: 17,
+    transform: [{ rotate: '45deg' }],
+    width: 17,
+  },
+});
+
+const getComputedStyles = (
```
RobinCsl: @jvaclavik Is it a standard way of solving this problem of dynamic styling? If that's performant and taking advantage of `StyleSheet.create`, then I'm all for it!

src/index.js
```diff
@@ -14,6 +14,7 @@ export { ServiceLogo } from './ServiceLogo';
 export { Stepper } from './Stepper';
 export { default as Text } from './Text';
 export { TextInput } from './TextInput';
+export { Tooltip, TooltipBubble } from './Tooltip';
```
RobinCsl: Is TooltipBubble necessary to export? It seems to me it is an internal component.

jvaclavik:
@RobinCsl I simplified measure logic in `Tooltip.js`, so it should be better optimised now. Could you check the component once again please?

src/Tooltip/Tooltip.js
tbergq: With this approach, the styles will be cached inside of the render method of the Tooltip component. The cached style is passed down, and if a child component re-renders, it will use the cached value. If tooltip re-renders, it will create a new cached value. 

I would rather split this into something like

```
const styles = StyleSheet.create({
  tooltipComponent: {
    position: 'relative',
    zIndex: 100,
  },
tooltipBubble: {
      position: 'absolute',
      zIndex: 101,
      opacity: 1,
    },
bubbleArrowDown: {top: 0},
bubbleArrowUp: {bottom: 0},
});
```

then using `style={[styles.tooltipBubble, isArrowDown ? styles.bubbleArrowDown : styles.bubbleArrowUp]}`

jvaclavik: @tbergq That make sense, but what if it's more complicated and I need to pass the argument to the style and decide which style key should be changed?

Here is the example:
```jsx
<View
  style={[
    styles.invisibleContainer,
    {
      [arrowVerticalPosition === 'flex-start'
        ? 'marginTop'
        : 'marginBottom']: verticalShift,
    },
  ]}
>
```
It's messy to have have those inline styles, so it make sense to separate style logic outside the JSX. 


What do you think about following solution?
```jsx
<View
  style={[
    styles.invisibleContainer,
    arrowVerticalPosition === 'flex-start'
      ? getComputedStyles(verticalShift).invisibleContainerTopShift
      : getComputedStyles(verticalShift).invisibleContainerBottomShift,
    style,
  ]}
>
```
```js
const getComputedStyles = verticalShift => StyleSheet.create({
  invisibleContainerTopShift: {
    marginTop: verticalShift,
  },
  invisibleContainerBottomShift: {
    marginBottom: verticalShift,
  },
});

```

Btw I'm not sure if it's necessary to use `StyleSheet.create` when I calculate styles depending on some parameters.

RobinCsl: Well, you could still put the logic of deciding on the correct style in the `render` method before the `return` statement. That means you can create all your styles in advance, have the logic in the render method to decide which one to use, and then if some are too dynamic to have been created in advance, you just add a dynamic property. WDYT?

src/Tooltip/TooltipBubble.js
RobinCsl: But I think this is the point of using `StyleSheet.create`, it sort of does that for you as far as I know. Apparently, you cannot notice the difference at first, but if an entire codebase did styles without it, it would become a huge burden on resources.

src/index.js
RobinCsl: Ok, makes sense.

src/Tooltip/Tooltip.js
jvaclavik: That sounds better, I will do it that way.

src/Tooltip/TooltipBubble.js
```diff
@@ -0,0 +1,93 @@
...
+    transform: [{ rotate: '45deg' }],
+    width: 17,
+  },
+  arrowContainer: {
+    flex: 1,
+    height: '100%',
+    paddingHorizontal: 15,
+    position: 'absolute',
+    width: '100%',
+    zIndex: 100,
```
RobinCsl: I believe there are also some defaultTokens for z-indexes. Would you mind having a look?

src/Tooltip/Tooltip.js
tbergq: On a general basis, if there are few options, use cached styles. If it is many I am ok with using inline styles. But it should be kept to a minimum 😊

src/Tooltip/TooltipBubble.js
jvaclavik: Done

src/Tooltip/TooltipBubble.js
```diff
@@ -0,0 +1,94 @@
...
+      </View>
+    </View>
+  );
+}
+
+const styles = StyleSheet.create({
+  text: {
+    color: defaultTokens.colorTextWhite,
+    padding: parseFloat(defaultTokens.spaceXSmall),
+    zIndex: 1,
```
RobinCsl: Sorry @jvaclavik, there is still one zIndex here.

RobinCsl:
Closing as PR #129 is the same and was merged (CI tests could run there)

PR #122, opened 2018-12-13
PlaceCard component

lukewalczak:


src/PlaceCard/Image/NetworkImage.js
```diff
@@ -0,0 +1,33 @@
...
+  if (!imageProps.source.uri) {
+    return <MissingImage {...imageProps} resizeMode="stretch" />;
+  }
+
+  const newProps = {
+    ...imageProps,
+    style: [
+      imageProps.style,
+      {
+        overflow: 'hidden', // otherwise 'borderRadius' won't work
```
RobinCsl: Would it make sense to have a StyleSheet.create for this style too?

src/PlaceCard/PlaceCard.js
```diff
@@ -0,0 +1,105 @@
...
+    ...StyleSheet.absoluteFillObject,
+    width: '100%',
+    borderRadius: 4,
+  },
+  badge: {
+    backgroundColor: defaultTokens.paletteProductNormal,
+    borderRadius: 4,
+  },
+  badgeText: {
+    fontSize: 16,
```
RobinCsl: Are there some defaultTokens we could use here?

PR #131, opened 2018-12-17
General: update react-native-web (0.9.9)

FilipMessa:


ferrannp:
Pity it seems they don't have a changelog and we need to check thru commits?

PR #134, opened 2018-12-17
fix: Use data-test-id instead of testID on web

### LINKED ISSUE #120: React does not recognize the `testID` prop on a DOM element
web button component should use data-test-id and not testID

ferrannp:
Closes #120.

This is ok. Just remember to use `data-test-id` on web. 

We could also write a babel plugin to enforce this or even easier... We can use https://www.npmjs.com/package/babel-plugin-jsx-remove-data-test-id and https://www.npmjs.com/package/babel-plugin-react-test-id to remove those attributes from production build.

If you would like any of the above, we can open an issue with it.

### DIFF
--- a/src/Button/Button.web.js
+++ b/src/Button/Button.web.js
@@ -46,8 +46,8 @@ export default function Button({
   if (href) {
     return (
       <a
-        testID={testID}
         href={href}
+        data-test-id={testID}
         style={{
           ...wrapperStyle,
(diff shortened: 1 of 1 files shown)


RobinCsl:
I think these plugins are definitely useful on the web to improve performance and shave off a few B of your bundle. We can look into them. I don't think it hurts to add them?

PR #135, opened 2018-12-17
Add script to generate flow for icon names

### LINKED ISSUE #128: Icon name flow type as enum
Would it be possible to generate the Icon flow-type name based on the icons.json file?

So we could get an error compile time if we select wrong icon name?

ferrannp:
Fixes #128.

- [x] Create script
- [x] Generated types should be part of `lib` too

### DIFF
--- a/.flowconfig
+++ b/.flowconfig
@@ -35,7 +35,6 @@
 node_modules/react-native/Libraries/react-native/react-native-interface.js
 node_modules/react-native/flow/
 node_modules/react-native/flow-github/
-scripts/types/index.flow.js
 
 [options]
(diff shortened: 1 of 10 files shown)


FilipMessa:
I was thinking about some basic explanation about why is this script here and what is it doing.
And where? probably in the readme.

ferrannp:
Ok, I added a comment to the script file. I do not think we need it in README as this is internal logic, consumer does not need to know or do anything about it (they will just get the right types when using the library).

FilipMessa:
Sure, make sense, sorry if I am annoying.

Issue #92, opened 2018-12-03
Create the Booking component 


FilipMessa:
# Booking Component

design: https://drive.google.com/drive/u/1/folders/1FvfYLKao8F2IZ0F1bmEkowIxCjRr-hWs

![05-results-list 1](https://user-images.githubusercontent.com/6354326/49371390-a0354200-f6f7-11e8-9984-1242669590d1.PNG)

FilipMessa:
Blocked, because the design will be updated

FilipMessa:
Design updated

FilipMessa:
You can start on this when issues #88 and #93 will be done.

PR #140, opened 2018-12-18
Add ConnectionCard component (#92)

jvaclavik:
screenshot 2018-12-18 at 14 41 52


src/ConnectionCard/BadgesContainer.js
```diff
@@ -0,0 +1,21 @@
+// @flow
+
+import * as React from 'react';
+import { ScrollView } from 'react-native';
+import StyleSheet from '../PlatformStyleSheet';
+
+type Props = {|
+  children: React.Node,
```
FilipMessa: pls add `read-only` for prop types https://flow.org/en/docs/types/interfaces/#toc-interface-property-variance-read-only-and-write-only

src/ConnectionCard/ConnectionCard.js
```diff
@@ -0,0 +1,102 @@
...
+type BadgeWithId = { ...BadgeProps, id: number };
+
+type Props = {|
+  padding?: boolean,
+  wayForth: Array<TripSectorWithId>,
+  wayBack?: Array<TripSectorWithId>,
+  price: PriceProps,
+  duration?: string,
+  badges?: Array<BadgeWithId>,
+|};
```
FilipMessa: read-only https://flow.org/en/docs/types/interfaces/#toc-interface-property-variance-read-only-and-write-only

src/ConnectionCard/ConnectionCardRow.js
```diff
@@ -0,0 +1,20 @@
...
+
+import * as React from 'react';
+import { View } from 'react-native';
+import StyleSheet from '../PlatformStyleSheet';
+import type { StylePropType } from '../PlatformStyleSheet/StyleTypes';
+
+type Props = {|
+  children: React.Node,
+  style?: StylePropType,
+|};
```
FilipMessa: read-only https://flow.org/en/docs/types/interfaces/#toc-interface-property-variance-read-only-and-write-only

src/ConnectionCard/ConnectionCardTypes.js
```diff
@@ -0,0 +1,13 @@
...
+
+export type TripSectorProps = {|
+  arrival: string,
+  arrivalTime: string,
+  carrier: CarrierData,
+  tripDate: string,
+  departure: string,
+  departureTime: string,
+  duration: string,
+|};
```
FilipMessa: read-only

src/Price/PriceTypes.js
```diff
@@ -0,0 +1,7 @@
+// @flow
+
+export type PriceProps = {|
+  value: number,
+  currency: string,
+  locale: string,
+|};
```
FilipMessa: read-only

src/ConnectionCard/ConnectionCard.js
```diff
@@ -0,0 +1,102 @@
...
+      arrivalTime={connection.arrivalTime}
+      carrier={connection.carrier}
+      tripDate={connection.tripDate}
+      departure={connection.departure}
+      departureTime={connection.departureTime}
+      duration={connection.duration}
+    />
+  ));
+
+const renderBadges = (badges: Array<BadgeWithId>) =>
```
FilipMessa: We don't need this prams type notation, because `renderBadges ` is not exported so flow is able to check this types without notation, but if this function has been exported you must add the notation to param.

src/ConnectionCard/ConnectionCard.js
```diff
@@ -0,0 +1,102 @@
...
+            </ConnectionCardRow>
+          )}
+          <ConnectionCardRow>
+            {wayBack && renderTripSector(wayBack)}
+          </ConnectionCardRow>
+        </React.Fragment>
+      )}
+      <ConnectionCardRow style={styles.lastRow}>
+        {hasBadges && (
+          <BadgesContainer>{badges && renderBadges(badges)}</BadgesContainer>
```
FilipMessa: I thinking about this part of the code
```
 <BadgesContainer>{badges && renderBadges(badges)}</BadgesContainer>
```

maby for better readability, we can do something like this in the `renderBadges` function:
(same for`renderTripSector` )
```
const renderBadges = badges => {
  if (!badges) return undefined;
   badges.map(badge => (
....
```
So main render code will be cleaner.
What do you think about that?

FilipMessa:
Please add the Price component to the storybook also.

FilipMessa:
And one more thing I look into the kiwi [glossary](https://kiwi.wiki/kiwi/handbook/#/comms/glossary) if ConnectionCard is the right name, and I am not sure now. 

> **Connection** - also called connecting flight is a flight that requires you to change from one plane to another at an intermediate airport (connection point). Connection consist of two or more flight segments. These segments are operated by one airline or airline group and later are booked under one PNR.

For me, it looks more like the itinerary.
https://kiwi.wiki/kiwi/handbook/#/comms/glossary?id=itinerary-parts-examples

What do you think?

src/ConnectionCard/ConnectionCard.js
jvaclavik: It's definitely more readable. And what about this one? I think it's even easier.

```js
const renderBadges = badges =>
  badges &&
  badges.map(badge => (
    <Badge key={badge.id} type={badge.type} style={styles.badge}>
      {badge.children}
    </Badge>
  ));

...

<BadgesContainer>{renderBadges(badges)}</BadgesContainer>
```

FilipMessa:
> And one more thing I look into the kiwi [glossary](https://kiwi.wiki/kiwi/handbook/#/comms/glossary) if ConnectionCard is the right name, and I am not sure now.
> 
> > **Connection** - also called connecting flight is a flight that requires you to change from one plane to another at an intermediate airport (connection point). Connection consist of two or more flight segments. These segments are operated by one airline or airline group and later are booked under one PNR.
> 
> For me, it looks more like the itinerary.
> https://kiwi.wiki/kiwi/handbook/#/comms/glossary?id=itinerary-parts-examples
> 
> What do you think?

After our discussion, and double check glossary I agree with you it is the connection.

src/ConnectionCard/ConnectionCard.js
FilipMessa: For me are both OK, so choose one.

src/Price/Price.stories.js
```diff
@@ -0,0 +1,19 @@
+// @flow
+
+import React from 'react';
+import { storiesOf } from '@storybook/react-native';
+import { withKnobs, number, select } from '@storybook/addon-knobs';
+import { Price } from './index';
+
+storiesOf('Price', module)
+  .addDecorator(withKnobs)
+  .add('Default', () => {
```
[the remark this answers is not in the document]
jvaclavik: What exactly is the difference between Playground and Default? Should be the Default without knobs?

jvaclavik:
Everything should be fixed now

src/Price/Price.stories.js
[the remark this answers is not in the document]
FilipMessa: I suppose that how you explained it is right.
It is what I saw across the universal-components's codebase.

PR #141, opened 2018-12-18
Add info on how to work with native storybook

### LINKED ISSUE #127: Make Storybook debugging easier on mobile device
If you run storybook on mobile device, there is a problem with time-consuming reload. When you update the component, you have to reload the page (or use live reload), open navigator and find your changed component. This is very ineffective especially for styling or debugging. 

I think there are following possible solutions:

- Use Hot reload – it doesn't work for now, but it would be probably the cleanest solution.
- Set custom initial screen in Storybook
- Possibility to set order of the components in navigator according to name (right now it's probably according to file loading sequence)

ferrannp:
I think this closes #127.


FilipMessa:
This Documentation is confusing, for me. 
It looks like you should run `yarn storybook-native` and then `yarn ios` or `yarn android`, but this doesn't work. As I figure it out. 

After `yarn storybook-native` you should run `react-native run-<platform>` if you want to benefit from `localhost:7007`

JoseDuSV:
> This Documentation is confusing, for me.
> It looks like you should run `yarn storybook-native` and then `yarn ios` or `yarn android`, but this doesn't work. As I figure it out.
> 
> After `yarn storybook-native` you should run `react-native run-<platform>` if you want to benefit from `localhost:7007`

Running `yarn storybook-native` then `yarn <platform>` works for me. In a same way as running `yarn storybook-native` => `react-native run-<platform>`. I only needed to restart Metro Bundler before 
successfully running it for first time.

README.md
```diff
@@ -20,12 +20,13 @@ yarn storybook
 
 ### Mobile
 
-After running `yarn`, you also need to link assets using:
-
 ```bash
-react-native link
+yarn storybook-native
```
JoseDuSV: Maybe we can make slightly more clear that this step is optional like in official docs https://github.com/storybooks/storybook/tree/next/app/react-native#start-storybook-server-optional .

Also for `Android` we should mention that sometimes port forwarding by running command `adb reverse tcp:7007 tcp:7007` is required to make connection to server work.

FilipMessa:
OK probably  I should restart Metro Bundler before
 running it for the first time. As @JoseDuSV mention

README.md
```diff
@@ -20,12 +20,14 @@ yarn storybook
 
 ### Mobile
 
-After running `yarn`, you also need to link assets using:
```
RobinCsl: Is this information somewhere else? Or is linking not necessary at all?

ferrannp: Linking for https://github.com/kiwicom/universal-components/blob/master/package.json#L96-L100.

README.md
RobinCsl: @JoseDuSV Can you add a commit to this PR with what you suggest?

RobinCsl:
Closing this as I expanded that section inside `CONTRIBUTING.md`, see #159

Issue #89, opened 2018-11-30
Create DatePicker component


RobinCsl:
There is a suggestion here of how we could do it

https://github.com/necolas/react-native-web/issues/1101

PR #143, opened 2018-12-19
Added: DatePicker component

lukewalczak:
Summary: For mobile we are using date pickers from `react-native` based on the native date pickers. For web we are going to use `react-datepicker` library and adjust it a bit to our requirements using prop `customInput`.

Fixes: #89

### DIFF
--- a/package.json
+++ b/package.json
@@ -33,6 +33,7 @@
   "dependencies": {
     "@kiwicom/orbit-design-tokens": "^0.2.5",
     "react-native-shimmer": "^0.4.1",
+    "react-datepicker": "2.0.0",
     "react-native-status-bar-height": "^2.2.0",
     "styled-components": "^4.1.2"
   },
--- a/.flowconfig
+++ b/.flowconfig
@@ -65,6 +65,8 @@ module.file_ext=.js
 module.file_ext=.jsx
 module.file_ext=.json
 module.file_ext=.native.js
+module.file_ext=.android.js
+module.file_ext=.ios.js
 
 suppress_type=$FlowIssue
 suppress_type=$FlowFixMe
--- a/src/DatePicker/DatePicker.android.js
+++ b/src/DatePicker/DatePicker.android.js
@@ -1,0 +1,140 @@
+// @flow
+
+import * as React from 'react';
+import { DatePickerAndroid, TimePickerAndroid } from 'react-native';
+
+import type { Props } from './DatePickerTypes';
+
+export default class AndroidDatePicker extends React.Component<Props> {
+  componentDidUpdate = (prevProps: Props) => {
+    const { isVisible, mode } = this.props;
+
+    if (!prevProps.isVisible && isVisible) {
+      if (mode === 'date' || mode === 'datetime') {
+        this.showDatePicker();
+      } else {
+        this.showTimePicker();
+      }
+    }
+  };
+
+  componentDidMount = () => {
+    const { isVisible, mode } = this.props;
+
+    if (this.props && isVisible) {
+      if (mode === 'date' || mode === 'datetime') {
+        this.showDatePicker();
+      } else {
+        this.showTimePicker();
+      }
+    }
+  };
+
+  showDatePicker = async () => {
+    const {
+      minDate,
+      maxDate,
+      datePickerMode,
+      mode,
+      date = new Date(),
+      onConfirm,
+      onDismiss,
+    } = this.props;
+    let picker;
+    try {
+      picker = await DatePickerAndroid.open({
+        date,
+        minDate,
+        maxDate,
+        mode: datePickerMode,
+      });
+    } catch (e) {
+      console.warn('Cannot open date picker', e.message);
+      return;
+    }
+
+    const { action, year, month, day } = picker;
+    // $FlowFixMe - it has a reference to DatePickerAndroid.ios.js
+    if (action !== DatePickerAndroid.dismissedAction) {
+      let customDate;
+      if (date && !isNaN(date.getTime())) {
+        const hour = date.getHours();
+        const minute = date.getMinutes();
+        customDate = new Date(year, month, day, hour, minute);
+      } else {
+        customDate = new Date(year, month, day);
+      }
+
+      if (mode === 'datetime') {
+        const timeOptions = {
+          mode: datePickerMode,
+          hour: date && date.getHours(),
+          minute: date && date.getMinutes(),
+        };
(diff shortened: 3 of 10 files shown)


JoseDuSV:
iOS and Android version work for me as we need for demo purpose.
And it's awesome that we don't need any extra dependency for it.

But I'm currently not sure what to do with web version. I'm not that happy about extra dependencies and webpack configuration mostly just because of Safari support. But I can live with it :)

My biggest concern is in terms of implementation. Because now there is no way how can I at least style generated input.

But I see in docs that `react-datepicker` can accept custom component as `customInput` prop.
Will it be possible to enable passing of this prop for web version, so I can use it like below?
```
<DatePicker
   onDismiss={this.hideDatePicker}
   onConfirm={this.handleDateChange}
   mode="date"
   customInput={
      <View>
         <Icon name="calendar" />
         <Text>Date: XYZ</Text>
      </View>
   }
/>
```

lukewalczak:
@JoseDuSV 

We have to create a reference to `DatePicker`

```
  refToDatePicker = (ref: ?DatePicker) => {
    this.calendar = ref;
  };
```

and then on `Button` in custom input, we have to call `setOpen(true)` on ref

```
      <DatePicker
        ref={this.refToDatePicker}
        selected={date}
        onChange={this.handleConfirm}
        minDate={minDate}
        maxDate={maxDate}
        customInput={
          <Button onPress={() => this.calendar.setOpen(true)}>
            Open date picker
          </Button>
        }
      />
```

I will add it tomorrow

lukewalczak:
ezgif com-video-to-gif 8

JoseDuSV:
Thanks, that is close. But I still ideally need some way how to set look of this button.
I think we can skip `ref` part because it seems that it is needed only if `Button` is used.
I've played with it and all I really need for **web** right now to be able to use it is to pass any element in props and then `DatePicker` will wrap around it.

So something like:
```
// index.web.js
.
.
<DatePicker
   selected={date}
   onChange={this.handleConfirm}
   minDate={minDate}
   maxDate={maxDate}
   customInput={this.props.customInput}
/>
.

// DatePicker.stories.js (or any other implementation)
.
.
<DatePicker
   isVisible={isOpen}
   onDismiss={this.hideDatePicker}
   onConfirm={this.handleDateChange}
   mode="date"
   customInput={
      <View style={{ flexDirection: 'row', padding: 10, backgroundColor: '#cccccc' }} >
      <Icon name="calendar" />
      <Text>Date: XYZ</Text>
      </View>
   }
/>
.
```

And then it will work/look like:

date-pick

src/DatePicker/component/index.ios.js
```diff
@@ -0,0 +1,137 @@
...
+export default class iOSDatePicker extends React.Component<Props, State> {
+  constructor(props: Props) {
+    super(props);
+
+    this.state = {
+      date: props.date || new Date(),
+    };
+  }
+
+  componentWillReceiveProps(nextProps: Props) {
```
JoseDuSV: Hopefully this component will be replaced until this method will be deprecated in version 17.
But just to be sure, can we at least name it `UNSAFE_componentWillReceiveProps` or use `getDerivedStateFromProps` ?

lukewalczak: I will refactor it to use `getDerivedStateFromProps`

src/DatePicker/component/index.web.js
```diff
@@ -0,0 +1,53 @@
...
+
+  handleConfirm = (date: Date) => {
+    const { onConfirm } = this.props;
+
+    this.setState({
+      date,
+    });
+
+    onConfirm(date);
+    this.setDefaultDate();
```
JoseDuSV: What about removing this reset call? I think that for web more common behaviour is to let picker stay on last selected day.

package.json
```diff
@@ -33,6 +33,8 @@
     "@kiwicom/orbit-components": "^0.23.0",
     "@kiwicom/orbit-design-tokens": "^0.2.5",
     "react-native-shimmer": "^0.4.1",
+    "intl": "^1.2.5",
```
FilipMessa: This should be removed, right? @RobinCsl

RobinCsl: That's right, a component library should not be concerned with l10n and i18n apart from RTL support, in my opinion.

RobinCsl: @lukewalczak Could you please remove `intl` library?

lukewalczak: I thought I removed it :( removing right now

src/DatePicker/component/DatePickerTypes.js
```diff
@@ -0,0 +1,15 @@
...
+export type Props = {|
+  +isVisible: boolean, // this prop is supported only on mobile
+  +mode?: 'date' | 'time' | 'datetime',
+  +datePickerMode?: 'calendar' | 'spinner' | 'default', // this prop is supported only on android
+  +date?: Date,
+  +minDate?: Date,
+  +maxDate?: Date,
+  +onConfirm: Date => void,
+  +onDismiss: () => void,
+  +customInput?: React.Node, // this prop is supported only on web
```
RobinCsl: I am just wondering here if it would make sense to do the following:

```
type Props = {|
  +mobile?: {|
       +isVisible: boolean, // this prop is supported only on mobile
  |},
  +android?: {|
     +datePickerMode?: 'calendar' | 'spinner' | 'default', // this prop is supported only on android
   |},
  +web?: {|
     +customInput?: React.Node, // this prop is supported only on web
   |},
  +mode?: 'date' | 'time' | 'datetime',
  +date?: Date,
  +minDate?: Date,
  +maxDate?: Date,
  +onConfirm: Date => void,
  +onDismiss: () => void,
|};
```

so that it's clearer to the consumer which props are accessible on which platform?
@tbergq Thoughts?

tbergq: No, I don't think we should do `mobile` `web` `android` it is ok to have some optional props just valid for one platform. Even react-native has that. 

It should be documented somewhere somehow though.

tbergq: So what we need is some documents page like [orbit](https://orbit.kiwi/)

JoseDuSV:
@RobinCsl Thanks for noticing.

@lukewalczak It seems to be related to `yarn.lock` update. When I roll back to commit `cda95e1` or use `yarn.lock` from `master` it works again as expected.

screenshot 2019-01-04 at 13 12 05

Issue #105, opened 2018-12-11
PlacePicker component


RobinCsl:
The end goal is to have something like this on mobile. For web, we might to think of how this could work. 

figma designs: https://www.figma.com/file/NTlh7HaG4KE4vDCBIIwEYe3Z/Place-Picker?node-id=0%3A1

- [ ] Input field should display typed text, as well as Badges of selected destinations (cf. below)
- [ ] Input field should have a **X** button to reset it
- [ ] Create row option component @lukewalczak PR #146 
- [ ] Create row options aggregator
- [ ] PLUS icon should add item to the input field as Badge component (which should be dismissed if pressing on it)
- [ ] State of the component should be controlled/controllable by external context
- [ ] Inquire how `KeyboardAvoidingView` is handled by React Native Web

<img width="320" alt="place-picker-from-typing" src="https://user-images.githubusercontent.com/22741774/49796427-f2501600-fd3c-11e8-9dc4-3de9ec913a17.png">

FilipMessa:
Start working on the first subtask 
- [ ] Input field should display typed text, as well as Badges of selected destinations

RobinCsl:
@FilipMessa I believe this was done in #212, closing.

PR #146, opened 2018-12-21
Added: PlacePicker row component

lukewalczak:
Summary: That PR will add `RowOption` component for `PlacePicker`. Within that sub-component have been done:

* different types of left icon with `triangle` shape below
* different types of bottom border
* right plus button (without functionality)

Fixes: This is sub-component of #105 

image


src/RowOption/RowOption.js
```diff
@@ -0,0 +1,205 @@
...
+                      style={styles.subheader}
+                      numberOfLines={1}
+                    >
+                      {info}
+                    </Text>
+                  </View>
+                )}
+              </View>
+            </View>
+            <Touchable onPress={onAddPress} style={styles.plusButton}>
```
RobinCsl: Could this be extracted into a smaller component?

src/RowOption/RowOption.js
```diff
@@ -0,0 +1,205 @@
+// @flow
+
+import * as React from 'react';
+import { View, Platform } from 'react-native';
+import { defaultTokens } from '@kiwicom/orbit-design-tokens';
+
+import Text from '../Text';
+import { Icon } from '../Icon';
+import Touchable from '../Button/Touchable';
```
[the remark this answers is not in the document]
lukewalczak: Sure, will handle this in separate PR.

src/RowOption/RowOptionContainer.js
```diff
@@ -0,0 +1,95 @@
...
+  +border?: 'long' | 'short' | 'shaped',
+  +type: 'localization' | 'destination' | 'airplane' | 'bus' | 'train',
+  +header: string | React.Node,
+  +subheader: string | React.Node,
+  +info?: string | React.Node,
+  +onItemPress: () => void,
+  +onAddPress: () => void,
+|};
+
+export default function RowOptionContainer({
```
JoseDuSV: Can we please name this file and component just `RowOption` to make clear that it is the main one. 
Also we probably should export it with `index.js` like the rest to make it consistent.

src/RowOption/__tests__/index.test.js
```diff
@@ -0,0 +1,42 @@
...
+
+  it('should have dot separator once info prop is passed', () => {
+    expect(getByTestId('dot-separator')).toBeDefined();
+  });
+
+  it('should have triangle shape once border type is shaped', () => {
+    expect(getByTestId('triangle-shape')).toBeDefined();
+  });
+
+  it('should execute onItemPress function', () => {
```
JoseDuSV: Can we please add test for `onAddPress`.
Also we just discussed that we should probably also test if event was actually fired like:
```
it('should execute onItemPress function', () => {
  fireEvent(getByType(Touchable), 'press');
  expect(onPress).toHaveBeenCalled(); 
});
```
( `onPress` in example will probably be renamed to `onItemPress` after we add extra test for `onAddPress`)

src/RowOption/RowOptionContainer.js
lukewalczak: 1.Sure I can rename it. 
2. Basically I didn't export it, because I thought that `RowOption` component will be in the future a part of `PlacePicker` component and will be within `PlacePicker` folder, but I will add `index.js` for it.

src/RowOption/__tests__/index.test.js
lukewalczak: Sure!

PR #147, opened 2018-12-21
fix: rnpm assets should be from lib/fonts

ferrannp:


scripts/generateTypes.js
```diff
@@ -12,7 +12,7 @@ const path = require('path');
 // eslint-disable-next-line import/no-extraneous-dependencies
 const rimraf = require('rimraf');
 
-if (fs.existsSync('../src/Icon/icons.json')) {
+if (fs.existsSync(path.join(__dirname, '../src/Icon/icons.json'))) {
   const icons = require('../src/Icon/icons.json'); // eslint-disable-line global-require
```
RobinCsl: Do we also need __dirname here?

ferrannp: Don't we? How would you write it?

RobinCsl: Well, I guess since you're checking that `path.join(__dirname, '../src/Icon/icons.json')` exists, you would need to require the same path. Otherwise, it might still throw an error?

RobinCsl: I guess you're right :)

PR #163, opened 2018-12-28
Added: CHANGELOG

RobinCsl:


mrtnzlml:
@RobinCsl Do you really need the changelog when the project is under heavy development with version 0.0.7? Do not block yourself with many processes around. This is why we do it only from version 1.0.x and we don't even write it for patch updates. I would personally wait when the project is more mature so you can now iterate fast.

Also, I am not super experienced with changelogs but I believe you should write them for your users. For me, it's a way how to communicate with everyone interested. So info about new components is 👍 but the users don't care about these things:

- Contributing guidelides extracted from README into CONTRIBUTING
- Package description reflects this is not a PoC anymore
- README contains link to deployed storybook
- ...

If I would be a user of this application then I would be interested in changes that can affect me, in new features and similar. I don't care about what's happening with the project/repository itself. Have a look at some real examples: https://github.com/facebook/react/blob/master/CHANGELOG.md or https://github.com/facebook/flow/blob/master/Changelog.md 🙂 

Maybe it's because of the _"All notable changes to this project will be documented in this file."_ sentence from _keepachangelog_? I believe they are wrong with this. 😬Should be _"All notable changes affecting our users will be documented in this file."_

RobinCsl:
These are very valid points, so we can probably leave this aside for now.

And you touch on an important aspect I overlooked: who the the intended recipient of this changelog is. If it's for users, then we should only document things concerning their usage of the library. If it's for us, maintainers, then it should contain more of the history of choices made in the repo. 
On the latter, of course the git history gives clues as to what happened, but a digestible format could still be useful. Recording why some choices were made can be interesting for posterity.

RobinCsl:
That makes great sense. I created an issue for it https://github.com/kiwicom/universal-components/issues/166

PR #168, opened 2019-01-02
Added: TimelineFlightDetail component

RobinCsl:
screenshot 2019-01-02 at 15 27 53

screenshot_1546439240

simulator screen shot - iphone x - 2019-01-02 at 15 27 03

### DIFF
--- a/package.json
+++ b/package.json
@@ -61,6 +61,7 @@
     "eslint-plugin-flowtype": "^3.2.0",
     "eslint-plugin-prettier": "^3.0.0",
     "eslint-plugin-react-native": "^3.5.0",
+    "fast-check": "^1.9.1",
     "file-loader": "^3.0.1",
     "flow-bin": "0.78.0",
     "jest": "^23.6.0",
--- a/src/CarrierLogo/CarrierLogoTypes.js
+++ b/src/CarrierLogo/CarrierLogoTypes.js
@@ -1,10 +1,10 @@
 // @flow
 
-export type CarrierData = {
-  code: string,
-  name: string,
-  type?: 'airline' | 'bus' | 'train',
-};
+export type CarrierData = {|
+  +code: string,
+  +name: string,
+  +type?: 'airline' | 'bus' | 'train',
+|};
 
 type Size = 'small' | 'medium' | 'large';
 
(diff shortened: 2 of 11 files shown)


tbergq:
Is this not very specific for a component library? I would expect the component library to expose something like Accordion. 

Then the grey area is something like a card with a header. 
Then each row could be a component

Then margarita patches it all together?

We are using relay in margarita, which means every component should fetch what it needs, now we fetch everything on one level, and props drill it down?

Thoughts?

RobinCsl:
@tbergq Well, this is definitely a discussion to be had.

If you look at the [Times component library](http://components.thetimes.co.uk/?knob-Lead%20Asset=true&knob-Size%20of%20ad%20placeholder%3A=default&knob-Video=true&knob-Headline=true&knob-Flags=true&knob-Standfirst=true&knob-Template=mainstandard&knob-Button%20width%3A%20=200&knob-Button%20title%3A=submit&knob-Linked%20Byline=true&knob-Label=true&selectedKind=Pages%2FArticle&selectedStory=Article%20with%20template%20choice&full=0&addons=1&stories=1&panelRight=1&addonPanel=storybooks%2Fstorybook-addon-knobs), they have some very specific components composing all the other ones.

I absolutely agree that our current setup does not really fit with Relay's idea about fragments, etc.

Another point to discuss is: do we want this component library to be generic or specific to our needs?

This whole component could be moved to a components folder inside `margarita`, but that means there won't be many components left in here as all the recent ones like `PlaceCard`, etc. are very specific.

tbergq:
Well, basically we are building [this](https://kiwicom.github.io/orbit-components/?selectedKind=TripSegment&selectedStory=Default&full=0&addons=1&stories=1&panelRight=0&addonPanel=storybook%2Factions%2Factions-panel) right?

I think this approach is better, because you have one component in charge of hiding and expanding what is inside (accordion) and the children is what ever you want. 

So back to specific or generic. For now, we have 2 consumers, margarita and mobile. There could be more in the future. 

In case both margarita and mobile were to use this component, what if margarita wants to remove seat recline and show something else, while mobile wants to keep seat recline? It would be a mess.

RobinCsl:
This can now be closed: PR open on Margarita https://github.com/kiwicom/margarita/pull/102

PR #174, opened 2019-01-04
docs: Added convention to write flow types

RobinCsl:
Summary: CONTRIBUTING now contains guidelines on how to write flow types for components which have platform specific files; refactored the types for Checkbox component as an example of what is expected.


RobinCsl:
@JoseDuSV Exactly. I will amend the docs to make that precise.

PR #177, opened 2019-01-07
FEAT: Create TagsInput component

FilipMessa:


FilipMessa:
Finally, the MR is ready, it was tricky find way how to handle it with one code because of differences between web input and native input.
Especially on the web input element, is no easy change width base on a count of characters.

I created the hidden element (sizer) and based on it I count the input width. It is kind of hacky but I can't find a better solution. 

So pls check it properly I spent a lot of the time on it so maby I missed something. Thanks

src/TagsInput/TagsInput.js
```diff
@@ -0,0 +1,160 @@
...
+import { MIN_WIDTH } from './constants';
+
+type Props = {|
+  +label: string,
+  +onChangeText: (value: string) => void,
+  +placeholder: string,
+  +selected?: string[],
+  +disabled?: boolean,
+  +fontSize: number,
+  +onClickDelete?: () => void,
```
JoseDuSV: Maybe just unify this with button name to something like `onClear` or `onClearPress`.

src/TagsInput/components/Label.js
```diff
@@ -0,0 +1,24 @@
...
+import { Text } from '../../Text';
+import type { StylePropType } from '../../PlatformStyleSheet/StyleTypes';
+
+type Props = {|
+  +text: string,
+  +weight: 'normal' | 'bold',
+  +style?: StylePropType,
+|};
+
+class Label extends React.PureComponent<Props> {
```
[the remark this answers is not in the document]
FilipMessa: Thanks 👍

src/TagsInput/TagsInput.js
```diff
@@ -0,0 +1,160 @@
...
+            maxWidth={this.getWidth(MIN_WIDTH)}
+            onLayout={this.setInputWidth}
+            value={value}
+            placeholder={
+              selected?.length === 0 && placeholder ? placeholder : null
+            }
+            onChangeText={this.handleChange}
+          />
+        </View>
+        {this.renderClearButton()}
```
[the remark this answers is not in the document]
FilipMessa: 💯 🥇

src/TagsInput/TagsInput.js
FilipMessa: i prefer onClearPress

FilipMessa:
I realize that there is the issue in test 
screen shot 2019-01-11 at 14 05 38

FilipMessa:
Issue  with test fixed

ios/storybooknative.xcodeproj/project.pbxproj
```diff
@@ -1085,6 +1106,13 @@
 				5C04BEA7D09C4DC4B05BB661 /* Roboto-MediumItalic.ttf in Resources */,
 				90151CF9607749A28BB60FDD /* Roboto-Regular.ttf in Resources */,
 				F9A904A78F284EE6989903AF /* orbit-icons.ttf in Resources */,
+				D88FDFB084E3435482711AB1 /* orbit-icons.ttf in Resources */,
```
JoseDuSV: Do you think this is related to your recent tests fix update? I'm just wondering why all fonts were added to `xcode` project for a second time.

FilipMessa: I'll check it

src/Badge/Badge.js
```diff
@@ -13,10 +13,15 @@ export default function Badge({
   children,
   type = 'primary',
   style,
+  fontSize,
 }: BadgeProps) {
   return (
     <View style={[styles.wrapper, theme(type).wrapper, style]}>
-      <Text style={[styles.text, theme(type).text]}>{children}</Text>
+      <Text
+        style={[styles.text, theme(type).text, fontSize ? { fontSize } : {}]}
```
RobinCsl: I don't know if that would improve performance much, but I suppose that doing:
```
const dynamicStyle = StyleSheet.create({
   fontSize: {
       fontSize,
   },
});
```
would make sure that if existing rule for fontSize exists, it won't be recreated in memory, etc.

src/TagsInput/TagsInput.js
```diff
@@ -0,0 +1,151 @@
...
+|};
+
+type State = {
+  value: string,
+  containerWidth: ?number,
+  inputWidth: ?number,
+};
+
+class TagInput extends React.Component<Props, State> {
+  inputRef: any; // @TODO find a way how to type this properly, the issue with mixing react and react-native
```
RobinCsl: I think we are using react > v16.3, so that could be helpful: https://reactjs.org/docs/refs-and-the-dom.html#creating-refs

src/TagsInput/TagsInput.js
```diff
@@ -0,0 +1,151 @@
...
+  setRef = (input: React.Element<any>) => {
+    this.inputRef = input;
+  };
+
+  getWidth = (width: ?number) => {
+    const { containerWidth } = this.state;
+    if (containerWidth && width) {
+      return containerWidth - width;
+    }
+    return undefined;
```
RobinCsl: I've always been told that it was better to return `null` than `undefined`, as `null` meant value could not be set instead of being undefined.

src/TagsInput/TagsInput.js
```diff
@@ -0,0 +1,151 @@
...
+  };
+
+  setRef = (input: React.Element<any>) => {
+    this.inputRef = input;
+  };
+
+  getWidth = (width: ?number) => {
+    const { containerWidth } = this.state;
+    if (containerWidth && width) {
+      return containerWidth - width;
```
RobinCsl: What if you get a negative result?

src/TagsInput/TagsInput.js
```diff
@@ -0,0 +1,151 @@
...
+    }
+  };
+
+  render() {
+    const { selected, fontSize, label, placeholder, disabled } = this.props;
+    const { value, inputWidth } = this.state;
+
+    return (
+      <View style={styles.container}>
+        <Text weight="bold" style={(styles.label, { fontSize })}>
```
RobinCsl: Same remark for fontSize as above.

src/TagsInput/TagsInput.js
```diff
@@ -0,0 +1,151 @@
...
+          <InputField
+            setRef={this.setRef}
+            fontSize={fontSize}
+            disabled={disabled}
+            minWidth={MIN_WIDTH}
+            maxWidth={this.getWidth(MIN_WIDTH)}
+            onLayout={this.setInputWidth}
+            value={value}
+            placeholder={
+              selected?.length === 0 && placeholder ? placeholder : null
```
RobinCsl: Can this logic be extracted before return statement?

src/TagsInput/TagsInput.js
```diff
@@ -0,0 +1,151 @@
...
+            }
+            onChangeText={this.handleChange}
+          />
+        </View>
+        <DeleteButton
+          style={{ opacity: value.length > 0 ? 1 : 0 }}
```
RobinCsl: Can this logic be extracted above the return statement? Also, can the style be added to `styles` below? This is important for performance.

src/TagsInput/TagsInput.js
```diff
@@ -0,0 +1,151 @@
...
+          onPress={this.handleClear}
+          disabled={!value.length ?? disabled}
```
RobinCsl: Same remark.

src/TagsInput/TagsInput.js
```diff
@@ -0,0 +1,151 @@
...
+    marginStart: 4,
+    overflow: 'hidden',
+  },
+
+  label: {
+    color: defaultTokens.paletteInkDark,
+  },
+});
+
+export default TagInput;
```
RobinCsl: Just to be consistent with other components, can you put the export default where you define the class?

src/TagsInput/components/DeleteButton.js
```diff
@@ -0,0 +1,37 @@
...
+  +style: StylePropType,
+  +disabled?: boolean,
+|};
+
+class DeleteButton extends React.PureComponent<Props> {
+  render() {
+    const { onPress, disabled, style } = this.props;
+    return (
+      <Touchable
+        testID="clear-button"
```
RobinCsl: Just to be consistent with the naming of your component, could you change it to "delete-button"?

src/TagsInput/components/DeleteButton.js
```diff
@@ -0,0 +1,37 @@
...
+          name="close"
+          color={defaultTokens.colorIconSecondary}
+          size="small"
+        />
+      </Touchable>
+    );
+  }
+}
+
+export default DeleteButton;
```
RobinCsl: Same comment as for TagsInput.

src/TagsInput/components/InputField.js
```diff
@@ -0,0 +1,134 @@
...
+      componentWidth: props.minWidth ?? 'auto',
+    };
+  }
+
+  handleOnLayout = (event: ViewLayoutEvent) => {
+    const { onLayout, maxWidth, minWidth, fontSize } = this.props;
+    const { width: inputWidth } = event.nativeEvent.layout;
+
+    const getWidth = width => {
+      const withBuffer = width + fontSize * 1.63;
```
RobinCsl: Could you explain where this 1.63 come from?

src/TagsInput/components/InputField.js
```diff
@@ -0,0 +1,134 @@
...
+      if (typeof maxWidth !== 'string' && withBuffer > maxWidth) {
```
RobinCsl: instead of `typeof maxWidth !== 'string'`, don't you want instead to make sure it's a number, like `typeof maxWidth === 'number'`?

src/TagsInput/components/InputField.js
```diff
@@ -0,0 +1,134 @@
...
+        return maxWidth;
+      }
+      if (
+        (typeof minWidth !== 'string' && withBuffer < minWidth) ||
+        !withBuffer
```
RobinCsl: Why this condition? It looks like withBuffer should be truthy?

src/TagsInput/components/InputField.js
```diff
@@ -0,0 +1,134 @@
...
+    } = this.props;
+    const { componentWidth } = this.state;
+
+    return (
+      <View style={styles.container}>
+        <Text
+          onLayout={this.handleOnLayout}
+          style={[
+            styles.hiddenSizer,
+            {
```
RobinCsl: Same comment as for dynamic styles earlier.

src/TagsInput/components/TagsContainer.js
```diff
@@ -0,0 +1,68 @@
...
+  };
+
+  render() {
+    const { tags, fontSize, maxWidth } = this.props;
+    if (!tags || tags.length < 1) {
+      return null;
+    }
+
+    return (
+      <View>
```
RobinCsl: Is this wrapping <View> necessary?

src/TagsInput/components/TagsContainer.js
```diff
@@ -0,0 +1,68 @@
...
+        <ScrollView
+          centerContent
+          horizontal
+          style={[styles.container, styles.row, { maxWidth }]}
+          showsHorizontalScrollIndicator={false}
+          contentContainerStyle={{
+            alignItems: 'center',
```
RobinCsl: Could you extract the style?

ios/storybooknative.xcodeproj/project.pbxproj
FilipMessa: I am working on it

src/Badge/Badge.js
FilipMessa: this should be inside the render?

FilipMessa: ah you mean this
```
const theme = (fontSize: number) =>
  StyleSheet.create({
   fontSize: {
       fontSize,
   },
  });
```

src/TagsInput/components/InputField.js
FilipMessa: 👍

src/TagsInput/components/InputField.js
FilipMessa: Sure this is from past when I was experimenting whit different approach.

src/TagsInput/components/TagsContainer.js
FilipMessa: It is, I need it for the web, if I don't use it break how overflow works

src/TagsInput/components/TagsContainer.js
FilipMessa: I tried before but I got back this flow error,
screen shot 2019-01-11 at 16 01 13

src/TagsInput/components/InputField.js
FilipMessa: This kind of magic, after experimenting with different value and this works best. I need this buffer because otherwise, it will be blink. Because it looks like the width of a character is a little bit different in TextInput against Text.

src/TagsInput/TagsInput.js
FilipMessa: OK, just there is another side of the coin, https://youtu.be/rhV6hlL_wMc?t=1326 😄 
But I will as you suggest.

src/TagsInput/TagsInput.js
FilipMessa: It can't happen I suppose, in the current use-case.
It can be zero but not negative.

src/TagsInput/TagsInput.js
FilipMessa: This was the first approach what I tried but, it didn't work.

src/TagsInput/components/InputField.js
RobinCsl: Would it be possible to include that explanation as a comment in the code?

src/TagsInput/components/TagsContainer.js
RobinCsl: Then we might need to address how our `StyleSheet.create` works, compared to RN's native one. Or use `StyleSheet.create` from RN implementation.
@tbergq What do you think about this issue?

src/TagsInput/TagsInput.js
RobinCsl: Thanks for the reference, I think he's got some very valid points. If we feel strongly about this, then we should enforce it.

src/TagsInput/components/InputField.js
FilipMessa: comment added

src/TagsInput/components/TagsContainer.js
FilipMessa: OK, I played with the flow for while but without success so I decided to use StyleSheet.create from RN implementation.

src/TagsInput/TagsInput.js
```diff
@@ -0,0 +1,162 @@
...
+  };
+
+  handleClear = () => {
+    const { onClearPress } = this.props;
+    this.setState({ value: '' });
+
+    // $FlowFixMe property focus is missing in object type
+    Platform.OS === 'web' && this.inputRef.current.focus();
+
+    if (onClearPress) {
```
RobinCsl: I believe you can even do `onClearPress?.()`.

src/TagsInput/TagsInput.js
```diff
@@ -0,0 +1,162 @@
...
+            maxWidth={this.getWidth(inputWidth)}
+            tags={selected}
+            fontSize={fontSize}
+          />
+          <InputField
+            ref={this.inputRef}
+            fontSize={fontSize}
+            disabled={disabled}
+            minWidth={MIN_WIDTH}
+            maxWidth={this.getWidth(MIN_WIDTH)}
```
RobinCsl: Maybe the naming could be improved here: why is `maxWidth` linked to `MIN_WIDTH`?

src/TagsInput/components/InputField.js
```diff
@@ -0,0 +1,135 @@
...
+  +maxWidth: ?number | ?string,
+  +minWidth: number,
+  +fontSize: number,
+  +disabled: ?boolean,
+  +placeholder: ?string,
+  +forwardRef: (el: any) => void,
+|};
+
+type State = {|
+  componentWidth: number | string,
```
RobinCsl:
```suggestion
  componentWidth: number | 'auto',
```

That's the only string that's allowed, right?

src/TagsInput/components/InputField.js
```diff
@@ -0,0 +1,135 @@
...
+  }
+}
+
+const styles = StyleSheet.create({
+  container: {
+    flexDirection: 'row',
+  },
+  input: {
+    web: {
+      outline: 'none',
```
RobinCsl: Can this be a problem for accessibility?

src/TagsInput/components/InputField.js
```diff
@@ -0,0 +1,135 @@
...
+    },
+  },
+  hiddenSizer: {
+    position: 'absolute',
+    flexDirection: 'row',
+    opacity: 0,
+    bottom: 0,
+    end: 0,
+    zIndex: -100,
```
RobinCsl: Would there be a design token for this zIndex?

src/TagsInput/TagsInput.js
FilipMessa: Because of in the getWidth method I subtract passing param, in this case, MIN_WIDTH from the width of the whole container.
And because TagsContainer minWidth is also MIN_WIDTH, maxWidth of the InputField must respect this. 

I will try to improve it to something more readable.

src/TagsInput/components/InputField.js
FilipMessa: I don't think that this zIndex should be in a token it is just an implementation detail no related to design.

src/TagsInput/components/InputField.js
FilipMessa: 👍

src/TagsInput/TagsInput.js
FilipMessa: thanks, I have been learned something new.

src/TagsInput/components/InputField.js
RobinCsl: You convinced me!

src/TagsInput/components/InputField.js
FilipMessa: I've improved website design. An inspiration I took from kiwi.com

src/TagsInput/components/InputField.js
RobinCsl: I cannot see it added now?

FilipMessa: fixed

PR #178, opened 2019-01-07
Docs: Added our commit template and instructions

RobinCsl:
Summary: This commit template is introduced to try and improve our commit message quality; the resulting git history will be more useful and documenting. Instructions on how to set it up were added to the contributing guidelines. (I also think that commit messages should start with a capital letter, so I modified the docs to reflect that.)


RobinCsl:
Yes, ideally it would come from one source of truth.

PR #182, opened 2019-01-08
Feat: Updated DatePicker

JoseDuSV:
Summary: Removed `react-datepicker` dependency which complicated implementation of component in `margarita`.

Replaced web component version with basic date picker implementation, which is closer to current mobile versions.

Small updates to iOS version to keep behavior consistent.

screenshot 2019-01-08 at 11 50 03

### DIFF
--- a/src/DatePicker/DatePicker.ios.js
+++ b/src/DatePicker/DatePicker.ios.js
@@ -19,44 +19,39 @@ type State = {
   date: Date,
 };
 
+const parsePropsToState = ({ date }: Props) => {
+  const tempDate = date ?? new Date();
+  return {
+    date: tempDate,
+  };
+};
+
 export default class iOSDatePicker extends React.Component<Props, State> {
   constructor(props: Props) {
     super(props);
 
     this.state = {
-      date: props.date || new Date(),
+      date: props.date ?? new Date(),
     };
   }
 
-  static getDerivedStateFromProps(nextProps: Props, nextState: State) {
-    const { date } = nextProps;
-
-    if (date && date !== nextState.date) {
-      return {
-        date,
-      };
+  componentDidUpdate = (prevProps: Props) => {
+    const { date } = this.props;
+    if (date != null && date !== prevProps.date) {
+      this.setState(parsePropsToState(this.props));
     }
-    return null;
-  }
-
-  setDefaultDate = () => {
-    this.setState({
-      date: new Date(),
-    });
   };
 
   handleDismiss = () => {
     const { onDismiss } = this.props;
-
     onDismiss();
+    this.setState(parsePropsToState(this.props));
   };
 
   handleConfirm = () => {
     const { onConfirm } = this.props;
     const { date } = this.state;
-
     onConfirm(date);
-    this.setDefaultDate();
   };
 
   handleChangeDate = (date: Date) => {
--- a/src/DatePicker/DatePicker.stories.js
+++ b/src/DatePicker/DatePicker.stories.js
@@ -4,23 +4,26 @@ import * as React from 'react';
 import { View, Platform } from 'react-native';
 import { storiesOf } from '@storybook/react-native';
 import { withKnobs, select, date as dateAddon } from '@storybook/addon-knobs';
-import { defaultTokens } from '@kiwicom/orbit-design-tokens';
+import { StyleSheet } from '../PlatformStyleSheet';
 
 import { Button } from '../Button';
-import { Icon } from '../Icon';
 import { Text } from '../Text';
 import DatePicker from './DatePicker';
-import { StyleSheet } from '../PlatformStyleSheet';
 
 type State = {
   isOpen: boolean,
-  date: string,
+  date: Date,
 };
+
+// NOTE: currentDate is used for default min & max limit values and is defined outside
+// of component because the default value for dateAddon must not change
+const currentDate = new Date();
+const dateMinMaxOffset = 1000 * 60 * 60 * 24 * 30;
 
 class DateTimePicker extends React.Component<{}, State> {
   state = {
     isOpen: false,
-    date: '',
+    date: new Date(),
   };
 
   showDatePicker = () => this.setState({ isOpen: true });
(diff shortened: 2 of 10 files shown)


src/DatePicker/DatePicker.web.js
```diff
@@ -1,52 +1,205 @@
...
+          <View style={styles.head}>
+            <Button
+              width={40}
+              onPress={this.swapMonthBack}
+              disabled={backDisabled}
+              type="secondary"
+              label="<"
+            />
+            <View style={styles.headLabel}>
+              <Text>{MONTHS[month]}</Text>
```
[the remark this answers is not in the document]
JoseDuSV: Definitely, thanks for noticing. I just added it to make this temporary version slightly nicer. 
But I can remove it or just replace it with month number for now.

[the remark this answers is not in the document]
JoseDuSV: Month name removed and replace just with number for now.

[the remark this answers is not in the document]
RobinCsl: We will still need some way of localising these. I guess we could create some LocalisationProvider component inside the component library that would be needed if you want to localise the lib and it would default to English. 
We're talking future improvement here. No need to do it now :)

src/DatePicker/DatePicker.ios.js
```diff
@@ -19,44 +19,37 @@ type State = {
   date: Date,
 };
 
+const parsePropsToState = date => {
```
RobinCsl:
```suggestion
const parsePropsToState = ({ date }: Props) => {
```

src/DatePicker/DatePicker.ios.js
```diff
@@ -19,44 +19,37 @@ type State = {
...
+};
+
 export default class iOSDatePicker extends React.Component<Props, State> {
   constructor(props: Props) {
     super(props);
 
-    this.state = {
-      date: props.date || new Date(),
-    };
+    this.state = parsePropsToState(props.date);
```
RobinCsl: What benefit is there to this approach compared to leaving it as it was before?

src/DatePicker/DatePicker.ios.js
```diff
@@ -19,44 +19,37 @@ type State = {
...
-  static getDerivedStateFromProps(nextProps: Props, nextState: State) {
-    const { date } = nextProps;
-
-    if (date && date !== nextState.date) {
-      return {
-        date,
-      };
+  componentDidUpdate = (prevProps: Props) => {
+    const { date } = this.props;
+    if (date && date !== prevProps.date) {
```
RobinCsl:
```suggestion
    if (date != null && date !== prevProps.date) {
```

src/DatePicker/DatePicker.web.js
```diff
@@ -1,52 +1,188 @@
...
+    const date = new Date(year, month, dayId);
+    this.setState(parsePropsToState(date));
+    onConfirm(date);
+  };
 
-    this.setState({
-      date,
+  swapMonthBack = () => {
+    this.setState(state => {
+      const month = state.month === 0 ? 11 : state.month - 1;
```
RobinCsl: Could these helper date functions be moved to a different file and can they be tested too?

src/DatePicker/DatePicker.web.js
```diff
@@ -1,52 +1,188 @@
...
-    const showTimeSelect = mode === 'time' || mode === 'datetime';
-    const showTimeSelectOnly = mode === 'time';
+    if (!isVisible) {
+      return null;
+    }
+
+    const monthLength = new Date(year, month + 1, 0).getDate();
+    const startDay = new Date(year, month, 1).getDay();
+    const selectedDay = new Date(
+      date.getFullYear(),
```
RobinCsl: I know that we recently had to use getUTCFullYear and the likes in `mobile` repo: https://github.com/kiwicom/mobile/commit/6d5e641de896e0b619775ea23f204a137d3a1a2a

JoseDuSV: Thanks, I'll look at it.

src/DatePicker/DatePicker.stories.js
```diff
@@ -40,7 +42,7 @@ class DateTimePicker extends React.Component<{}, State> {
     );
     const maxDateStringTimestamp = dateAddon(
       'Max date',
-      new Date('Jan 20 2019')
+      new Date('Feb 20 2019')
```
FilipMessa: what about changing this to something dynamic base on a current date?
Something like this, but there should be way how to write it nicer then my example.
```
      let today = new Date('Feb 20 2019');
 
      const maxDateStringTimestamp = dateAddon(
        'Max date',	 
        today.setMonth(today.getMonth() + 8);
  )
```

src/DatePicker/DatePicker.web.js
```diff
@@ -1,52 +1,188 @@
...
 
-    this.setState({
-      date,
+  swapMonthBack = () => {
+    this.setState(state => {
+      const month = state.month === 0 ? 11 : state.month - 1;
+      return {
+        month,
+        year: month === 11 ? state.year - 1 : state.year,
+      };
```
FilipMessa: for consistency will be better set month and year with the same approach.
```
const mount = ... ;
const year = ... ;
```

or

```
return {
   mount: state.mount === 0 ... ,
   year: month === 11 ? state.year - 1 : state.year,
}
```

src/DatePicker/DatePicker.web.js
```diff
@@ -1,52 +1,188 @@
...
     });
+  };
 
-    onConfirm(date);
+  swapMonthForward = () => {
+    this.setState(state => {
+      const month = (state.month + 1) % 12;
+      return {
+        month,
+        year: month === 0 ? state.year + 1 : state.year,
```
FilipMessa: same as 🔝, but it can be also as a helper function, like Robin mention before.

src/DatePicker/DatePicker.web.js
JoseDuSV: Moved and tests added.

src/DatePicker/DatePicker.stories.js
JoseDuSV: Good idea! Implemented. At the end I've used just fixed thirty days offset which is slightly shorter than using `setMonth` `getMonth` combo and the final effect is quite equal.

src/DatePicker/DatePicker.web.js
JoseDuSV: Definitely, updated.

src/DatePicker/DatePicker.web.js
JoseDuSV: Updated, extracted, tests added.

Issue #103, opened 2018-12-10
Create MultiSlider component


RobinCsl:
Using `react-native-multi-slider` was not enough for our needs (cf. #97 ). We now have more precise designs and documentation about what the component should be able to do.

https://paper.dropbox.com/doc/Slider-specification--ATbeMyF~KvJEOrDN6vJg8ZINAg-ZFQOT2Hyfd183tyGvmEoK

PR #184, opened 2019-01-08
Feat: Create Slider/Multislider component

lukewalczak:
That PR will add slider/multislider component

Fixes: #103

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -19,7 +19,7 @@ jobs:
             - v1-30-dependencies-
       - run:
           name: Install Dependencies
-          command: yarn
+          command: yarn --network-concurrency 1 install
       - save_cache:
           key: v1-30-dependencies-{{ checksum "package.json" }}
           paths:
--- a/.storybook/webpack.config.js
+++ b/.storybook/webpack.config.js
@@ -9,14 +9,23 @@ module.exports = {
           'react-native-web-image-loader?name=[name].[ext]&scalings[@2x]=2&scalings[-3x]=3',
       },
       {
+        test: /\.js$/,
+        include: [
+          path.join(__dirname, '../node_modules/react-native-multi-slider'),
+        ],
+        use: 'babel-loader',
+      },
+      {
         test: /\.(woff(2)?|ttf|eot|svg)(\?v=\d+\.\d+\.\d+)?$/,
-        use: [{
+        use: [
+          {
             loader: 'file-loader',
             options: {
-                name: '[name].[ext]',
-                outputPath: 'fonts/'
-            }
-        }]
+              name: '[name].[ext]',
+              outputPath: 'fonts/',
+            },
+          },
+        ],
       },
     ],
   },
--- a/.storybook/webpack.config.js
+++ b/.storybook/webpack.config.js
@@ -9,13 +9,6 @@ module.exports = {
           'react-native-web-image-loader?name=[name].[ext]&scalings[@2x]=2&scalings[-3x]=3',
       },
       {
-        test: /\.js$/,
-        include: [
-          path.join(__dirname, '../node_modules/react-native-multi-slider'),
-        ],
-        use: 'babel-loader',
-      },
-      {
         test: /\.(woff(2)?|ttf|eot|svg)(\?v=\d+\.\d+\.\d+)?$/,
         use: [
           {
--- a/package.json
+++ b/package.json
@@ -32,7 +32,7 @@
   "license": "MIT",
   "dependencies": {
     "@kiwicom/orbit-design-tokens": "^0.2.5",
-    "react-native-multi-slider": "https://github.com/kiwicom/react-native-multi-slider.git#a53e20f97d93361907d283dd98eb6da3c20c4249",
+    "react-native-multi-slider": "https://github.com/kiwicom/react-native-multi-slider.git",
     "react-native-status-bar-height": "^2.2.0",
     "styled-components": "^4.1.2"
   },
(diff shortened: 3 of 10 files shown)


.storybook/webpack.config.js
```diff
@@ -8,15 +8,24 @@ module.exports = {
         use:
           'react-native-web-image-loader?name=[name].[ext]&scalings[@2x]=2&scalings[-3x]=3',
       },
+      {
+        test: /\.js$/,
+        include: [
+          path.join(__dirname, '../node_modules/react-native-multi-slider'),
```
RobinCsl: I will need to evaluate whether this approach will work for us; I've been trying to reduce the amount of stuff that goes into webpack config files as just installing the library might not work.

package.json
```diff
@@ -33,6 +33,7 @@
   "dependencies": {
     "@kiwicom/orbit-design-tokens": "^0.2.5",
     "react-datepicker": "2.0.0",
+    "react-native-multi-slider": "https://github.com/lukewalczak/react-native-multi-slider.git#feat/block-overlap",
```
RobinCsl: Great work! We will need to create an issue to take over that fork.

lukewalczak: I can transfer ownership to you or you can create a fork from my fork

yarn.lock
```diff
@@ -9,18 +9,38 @@
   dependencies:
     "@babel/highlight" "^7.0.0"
 
+"@babel/core@7.1.0":
```
JoseDuSV: There are quite a lot of changes inside `yarn.lock`. Are you able to run it on `Android`?
Because for me the lockfile related problem that we solved few days ago is back.

screenshot 2019-01-09 at 15 47 09

So I'm just thinking if it's not the same case.

lukewalczak: It was working before, but I've rebased it with master later. I'm going to resolve conflicts and rebase it again.

JoseDuSV: Ok, thanks.

src/Slider/Slider.js
```diff
@@ -0,0 +1,263 @@
...
+import MultiSlider from 'react-native-multi-slider';
+
+import Marker from './Marker';
+import { Text } from '../Text';
+import { StyleSheet } from '../PlatformStyleSheet';
+
+const LABEL_LENGTH = 200;
+const TRACK_HEIGHT = 4;
+
+type Props = {|
```
JoseDuSV: Maybe I'm just blind. But shouldn't there also be something like `onValuesChange` between props so I can handle value/s changes?

lukewalczak: Nope, you're right - I will expose tomorrow following props: `onValuesChange`, `onValuesChangeStart`, `onValuesChangeFinish`.

src/Slider/Slider.js
```diff
@@ -0,0 +1,263 @@
...
+    });
+  };
+
+  calculateOffset = (value: number) => {
+    const { minValue, maxValue } = this.props;
+    const { width } = this.state;
+    return (value * width) / (maxValue - minValue) - LABEL_LENGTH / 2;
+  };
+
+  createParts = () => {
```
JoseDuSV: What about extracting this part into some separate `TrackLine` component? Maybe even with those current two `Views` with `trackLine` style.

JoseDuSV:
jan-09-2019 17-47-49

There is currently some short ugly jump in web version during initialisation. It will be nice to take care of it somehow or at least show component after everything is set. But it probably doesn't need to be solved right now and we can create separate issue for this improvement.

Otherwise awesome work!

.storybook/webpack.config.js
RobinCsl: If you try to install @kiwicom/universal-components@0.0.9-alpha.0 in a web project, it does not work, even after modifying the webpack config to load the files apparently.

https://github.com/kiwicom/margarita/tree/testing-slider-kiwicom-alpha-release

src/Slider/Slider.js
```diff
@@ -0,0 +1,198 @@
...
+  +maxValue: number,
+  +startValue: number,
+  +endValue?: number,
+  +label: string | React.Node,
+  +numOfParts?: number,
+  +snapped?: boolean,
+  +step?: number,
+  +customMarker?: React.Node,
+  +sliderLength?: number,
+  +onValuesChange?: () => void,
```
JoseDuSV: Can `onValuesChange` also return actual slider value?

For example on implementation below (and in current mobile app) the **Time of stay** value should change during dragging.

screenshot 2019-01-10 at 15 18 59

.circleci/config.yml
```diff
@@ -19,7 +19,7 @@ jobs:
             - v1-30-dependencies-
       - run:
           name: Install Dependencies
-          command: yarn
+          command: yarn --network-concurrency 1 install
```
RobinCsl: What does that do?

lukewalczak: it will set maximum number of concurrent network to one. Without `install dependencies` step was crashing in that PR

RobinCsl: I believe this is explained here why this could be necessary: https://github.com/yarnpkg/yarn/pull/2129

And it makes sense: we're fetching the code from your fork on GitHub right now.

.storybook/webpack.config.js
```diff
@@ -8,15 +8,24 @@ module.exports = {
         use:
           'react-native-web-image-loader?name=[name].[ext]&scalings[@2x]=2&scalings[-3x]=3',
       },
+      {
+        test: /\.js$/,
```
RobinCsl: Do you think we can remove this now, @lukewalczak ?
It should not be necessary, right?

lukewalczak: Right, now it's redundant.

PR #189, opened 2019-01-11
Fix: Unified DatePicker behavior

JoseDuSV:
Summary:
- Set default `mode` to `date` for iOS and Android.
- Fixed bug in `web` version, when `minDate` tile was disabled in cases where time of the day was set to different value than 0.

### DIFF
--- a/src/DatePicker/DatePicker.web.js
+++ b/src/DatePicker/DatePicker.web.js
@@ -7,7 +7,11 @@ import { StyleSheet } from '../PlatformStyleSheet';
 import DatePickerDayTile from './DatePickerDayTile';
 import { Button } from '../Button';
 import { Text } from '../Text';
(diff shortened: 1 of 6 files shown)


src/DatePicker/DatePicker.web.js
```diff
@@ -81,14 +85,12 @@ export default class WebDatePicker extends React.Component<Props, State> {
       const dayId = i + 1 - startDay;
       const tempDateTime =
         dayId > 0 ? new Date(year, month, dayId).getTime() : 0;
-      const disabled =
-        tempDateTime < (minDate ?? 0) ||
-        tempDateTime > (maxDate ?? Number.MAX_SAFE_INTEGER);
+      const disabled = tempDateTime < minDateTime || tempDateTime > maxDateTime;
```
RobinCsl: I like that a lot, this is much clearer.

PR #190, opened 2019-01-11
Docs: Added nicer documentation system, complementing Storybook

RobinCsl:
Summary: This documentation is generated by docz.site.
It needs a separate package.json in a different folder because there currently is a conflict with `@emotion` library versions as Storybook uses a different version.
I added some basic instructions on how to get started with example projects with CRA and Expo.
I also added an example of how the documentation system can be used: see Accordion.mdx.
If this solution is liked, I will create auto-deploy with Netlify too.

A preview can be seen here for now: https://wizardly-bassi-29ad3e.netlify.com/


index.mdx
```diff
@@ -0,0 +1,77 @@
...
+You could add the following snippet in `App.css`:
+```
+@font-face {
+  font-family: "orbit-icons";
+  src: url(../node_modules/@kiwicom/universal-components/lib/fonts/orbit-icons.ttf);
+}
+```
+
+>Two working examples on GitHub:    
+- [in the example folder](https://github.com/kiwicom/universal-components/tree/master/examples/create-react-app-icon-font) of the @kiwicom/universal-component repository,
```
JoseDuSV: Do you think that it needs an extra example? I mean, every example is definitely plus, but is it worth it in this case? If we consider that we will need to maintain and update this sub-project from time to time.
I think that implementation description itself is quite clear and simple.

RobinCsl: Yeah, I thought about that.
But depending on whether this repo joins Margarita or not in the future, I think people might want to look at this repo and if they see nice documentation with some examples, they might be a lot more willing to try it out and star it.
And maintaining these examples should not be a hard task (and if they ever fall out of date, people might raise an issue?)
I guess the point I was trying to make with these examples is they can clone the repo, and go into the examples folder and not touch anything else.

RobinCsl:
> Did you considered some other options or was docz first choice right away because of its features?

I looked at Docusaurus but it does not yet have mdx support.
I looked at how CallStack created their docs for react-native-paper.
The features offered by `docz` were quite convincing for me: just drop mdx files wherever, construct prop tables, offer a playground where people can see and copy/modify the code, and what's more, it's got React Native support through react-native-web out of the box. That was pretty compelling for me.

index.mdx
JoseDuSV: You convinced me. Having repo that I can clone and start playing with right away is definitely plus.

PR #191, opened 2019-01-14
Update Badge style

jvaclavik:
Graphic design for badges is different than our in `universal-components'. So that is the reason for this PR. 

In the future, we need to have orbit design tokens updated because there are different badge background/text colours and height. I sent those issues to orbit team.

screenshot 2019-01-14 at 17 23 46

### DIFF
--- a/src/Badge/Badge.js
+++ b/src/Badge/Badge.js
@@ -34,18 +34,17 @@ const styles = StyleSheet.create({
     alignSelf: 'flex-start',
     flexDirection: 'row',
     flexWrap: 'wrap',
-    borderRadius: parseFloat(defaultTokens.borderRadiusBadge),
-    paddingHorizontal: 8,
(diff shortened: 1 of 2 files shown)


JoseDuSV:
I don't think this is really the issue of orbit design tokens. And we can't probably expect that those values will be updated just for us ( if there isn't a plan to change it in all current products).
Maybe we can solve it by implementing themes like in Orbit https://orbit.kiwi/guidelines/theming/ or by extending defaultTokens in some other way.

JoseDuSV:
@jvaclavik Ok, as you stated on Slack. It will be definitely beneficial to have a talk with Orbit team and figure out some solution. And I can definitely imagine something similar to what you've proposed.
Maybe we should also invite someone from `mobile` team, to check how/if they are currently using design tokens.

Thanks!

PR #199, opened 2019-01-22
Fix: Text component cascades styles

### LINKED ISSUE #198: Support nested text
Now when doing this:

```
<Text style={{ color: 'blue', fontSize: 22 }}>
    Test
	<Text style={{ color: 'red' }}>Nested test</Text>
</Text>
```

I see 

screenshot 2019-01-22 at 11 10 59

But I expected 

screenshot 2019-01-22 at 11 12 21

RobinCsl:
Summary: To extend React Native's Text component, we should allow styles to cascade. This behaviour was prevented because default values ewere set for the size, the weight and the type.

Closes #198

### DIFF
--- a/src/Text/__tests__/__snapshots__/index.test.js.snap
+++ b/src/Text/__tests__/__snapshots__/index.test.js.snap
@@ -5,39 +5,33 @@ exports[`Text should match snapshot diff 1`] = `
 - First value
 + Second value
 
-@@ -3,24 +3,28 @@
+@@ -3,10 +3,28 @@
       Array [
         Object {
           \\"fontFamily\\": \\"Roboto\\",
           \\"margin\\": 0,
         },
--       undefined,
--       undefined,
-        Object {
--         \\"fontSize\\": 14,
++       Object {
 +         \\"fontStyle\\": \\"italic\\",
-        },
-        Object {
--         \\"fontSize\\": 14,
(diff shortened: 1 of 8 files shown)


RobinCsl:
To see the solution in action:
https://deploy-preview-199--kiwicom-universal-components.netlify.com/?selectedKind=Text&selectedStory=Cascading%20styles&full=0&addons=1&stories=1&panelRight=0&addonPanel=storybook%2Factions%2Factions-panel

src/Text/__tests__/__snapshots__/index.test.js.snap
```diff
@@ -5,39 +5,40 @@ exports[`Text should match snapshot diff 1`] = `
...
 +       Object {
 +         \\"textAlign\\": \\"center\\",
 +       },
         undefined,
+-       undefined,
+-       undefined,
+-       undefined,
+-       undefined,
+-       undefined,
+-       undefined,
```
FilipMessa: probably will be nice have there some values instead of `undefined`

RobinCsl: Okay, I will improve the implementation to avoid these `undefined` from popping up.

RobinCsl:
@FilipMessa It should be better now 🙂

PR #200, opened 2019-01-22
Fix default Badge's font size

jvaclavik:
Default font size in Badge was too big.

### DIFF
--- a/src/Badge/Badge.js
+++ b/src/Badge/Badge.js
@@ -15,11 +15,15 @@ export default function Badge({
   style,
   fontSize,
 }: BadgeProps) {
-  const dynamicStyle = StyleSheet.create({
-    fontSize: {
-      fontSize,
-    },
-  });
+  let dynamicStyle = { fontSize: {} };
+  if (fontSize != null) {
+    dynamicStyle = StyleSheet.create({
+      fontSize: {
+        fontSize,
+      },
+    });
+  }
+
   return (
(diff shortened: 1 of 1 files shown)


src/Badge/Badge.js
```diff
@@ -13,7 +13,7 @@ export default function Badge({
   children,
   type = 'primary',
   style,
-  fontSize,
+  fontSize = 12,
```
RobinCsl: I think a better fix is the following:
line 25, replace `dynamicStyle.fontSize` with `fontSize != null && dynamicStyle.fontSize`.
The issue is that we are applying
```
...
fontSize: undefined,
...
```
when fontSize is not given.

Alternatively, replace that default value with `parseFloat(defaultTokens.fontSizeTextSmall)`.

I would still want this dynamicStyle to not be used if fontSize is not specified, because it would be recreated for every re-render of the badge, where styles.text should live in memory and not be recreated on every re-render.

RobinCsl: So maybe something better would be:
```
let dynamicStyle = { fontSize: {}};
if (fontSize != null) {
   dynamicStyle = StyleSheet.create({
    fontSize: {
      fontSize,
    },
  });
}
```

and replace `styles.text.fontSize` by `parseFloat(defaultTokens.fontSizeTextSmall)`.

jvaclavik: Done

PR #203, opened 2019-01-22
Feat: Added SegmentedButton component

JoseDuSV:
Summary: Component allows selection from multiple options. Will be used for gender selector in passenger part of booking screen.

Related to:
https://github.com/kiwicom/margarita/issues/73
https://github.com/kiwicom/margarita/issues/121

screenshot 2019-01-22 at 10 54 42

### DIFF
--- a/src/SegmentedButton/Segment.js
+++ b/src/SegmentedButton/Segment.js
@@ -1,0 +1,83 @@
+// @flow
+
+import * as React from 'react';
+import { View, TouchableWithoutFeedback } from 'react-native';
+import { defaultTokens } from '@kiwicom/orbit-design-tokens';
+import { Text } from '../Text';
+import { StyleSheet } from '../PlatformStyleSheet';
+
+export type Props = {|
+  +label: string,
+  +value: string,
+  +first: boolean,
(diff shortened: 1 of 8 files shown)


src/SegmentedButton/Segment.js
```diff
@@ -0,0 +1,83 @@
...
+  };
+
+  render() {
+    const { label, first, last, active } = this.props;
+    return (
+      <TouchableWithoutFeedback onPress={this.handleOnPress}>
+        <View
+          style={[
+            styles.container,
+            first && styles.containerFirst,
```
RobinCsl: As I refactored `Text` in a previous PR, maybe it would be cleaner to remove all these undefined from `style` and creating a helper function.

JoseDuSV: I can definitely do that. But is there any added value in it, besides slightly nicer test snapshots? For me it's quite similar and we are using current `[styles.label, active && styles.labelActive]` pattern in multiple components. It is also mentioned in StyleSheet part of RN docs ( https://facebook.github.io/react-native/docs/stylesheet ).

Issue #201, opened 2019-01-22
Imported propTypes doesn't work with Docz PropsTable


JoseDuSV:
As mentioned in this issue:
https://github.com/pedronauck/docz/issues/129
it is probably related to `docgen`
https://github.com/reactjs/react-docgen/issues/33

![51674708-b95cbf80-1fd1-11e9-8584-1bd3f37ae3a0](https://user-images.githubusercontent.com/2660330/51674836-21aba100-1fd2-11e9-88ac-af04b41de059.png)

PR #204, opened 2019-01-22
Feat: Added Picker component

JoseDuSV:
Summary: Picker implementation for each platform. Will be used for example as nationality selector in passenger section of booking screen.

Related to:
https://github.com/kiwicom/margarita/issues/73
https://github.com/kiwicom/margarita/issues/121

Desktop:
desktop

iOS:
ios

Android:
android

### DIFF
--- a/src/Picker/Picker.ios.js
+++ b/src/Picker/Picker.ios.js
@@ -1,0 +1,139 @@
+// @flow
+
+import * as React from 'react';
+import { View } from 'react-native';
+import { defaultTokens } from '@kiwicom/orbit-design-tokens';
+
+import { Button } from '../Button';
+import { Icon } from '../Icon';
+import { Modal } from '../Modal';
+import { StyleSheet } from '../PlatformStyleSheet';
+import type { Props } from './PickerTypes';
+import { getSelectedLabel } from './PickerHelpers';
+import NativePicker from './NativePicker';
+
+export type State = {|
+  open: boolean,
+  selectedValue: string,
+  pickerValue: string,
+|};
+
+export default class Picker extends React.Component<Props, State> {
+  static defaultProps = {
+    placeholder: '',
+  };
+
+  constructor(props: Props) {
+    super(props);
+
+    const { selectedValue } = this.props;
+
+    this.state = {
+      open: false,
+      selectedValue,
+      pickerValue: selectedValue,
+    };
+  }
+
+  componentDidUpdate = (prevProps: Props) => {
+    if (
+      this.props.selectedValue !== prevProps.selectedValue &&
+      this.props.selectedValue !== this.state.selectedValue
+    ) {
+      this.setState({
+        selectedValue: this.props.selectedValue,
+        pickerValue: this.props.selectedValue,
+      });
+    }
+  };
+
+  handleOpenPress = () => {
+    this.setState({
+      open: true,
+    });
+  };
+
+  handleModalClose = () => {
+    this.setState({
+      open: false,
+    });
+  };
+
+  handlePickerValueChange = (value: string) => {
+    this.setState({ pickerValue: value });
+  };
+
+  handleOkPress = () => {
+    const { pickerValue } = this.state;
+    const { onValueChange } = this.props;
+    this.setState({
+      open: false,
+      selectedValue: pickerValue,
+    });
+    onValueChange(pickerValue);
+  };
+
+  render() {
+    const { open, selectedValue, pickerValue } = this.state;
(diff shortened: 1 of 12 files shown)


RobinCsl:
This is in the storybook preview. There might be a small issue. Could you look at it @JoseDuSV ?

screenshot 2019-01-23 at 10 21 00

JoseDuSV:
> This is in the storybook preview. There might be a small issue. Could you look at it @JoseDuSV ?
> 
> screenshot 2019-01-23 at 10 21 00

Thanks, it's fixed now.

JoseDuSV:
@RobinCsl Conflict solved.

And you are right, there is definitely room for improvement in web version. For now there is native select used in passenger form section even on kiwi (only button style is slightly different).

screenshot 2019-01-23 at 14 44 20

But we definitely don't need to strictly follow it and we can make the select items style nicer and closer to class selector on homepage.

screenshot 2019-01-23 at 14 44 45

Can we just merge this PR to make component available and I will create extra issue for web version improvements. Thanks

src/Picker/Picker.ios.js
```diff
@@ -0,0 +1,119 @@
...
+      open: false,
+      selectedValue: pickerValue,
+    });
+    onValueChange(pickerValue);
+  };
+
+  render() {
+    const { open, selectedValue, pickerValue } = this.state;
+    const { optionsData } = this.props;
+    const selectedOption = optionsData.find(
```
RobinCsl: This logic, and the corresponding component is duplicated between ios and android. Could you extract it?

src/Picker/Picker.ios.js
```diff
@@ -0,0 +1,119 @@
...
+        {open && (
+          <View style={styles.container}>
+            <RNPicker
+              selectedValue={pickerValue}
+              style={styles.picker}
+              onValueChange={this.handlePickerValueChange}
+            >
+              {pickerOptions}
+            </RNPicker>
+            <Button onPress={this.handleOkPress} type="secondary" label="OK" />
```
RobinCsl: I am not sure about this label. I guess OK is pretty international. But we really need to be wary of this. What could we do about it?

src/Picker/Picker.web.js
```diff
@@ -0,0 +1,89 @@
...
+
+import * as React from 'react';
+import { View } from 'react-native';
+import { defaultTokens } from '@kiwicom/orbit-design-tokens';
+import { Icon } from '../Icon';
+import { StyleSheet } from '../PlatformStyleSheet';
+
+import type { Props } from './PickerTypes';
+
+const DEFAULT_OPTION = 'default-empty-option';
```
RobinCsl: Could you explain what this achieves?

RobinCsl:
Also, this is not strictly related to this PR, but we will need to find a way to properly pass prop types to the components in MDX files as Docz does not parse them if they are outside.
screenshot 2019-01-24 at 12 14 29

JoseDuSV:
> Also, this is not strictly related to this PR, but we will need to find a way to properly pass prop types to the components in MDX files as Docz does not parse them if they are outside.
> screenshot 2019-01-24 at 12 14 29

@RobinCsl  Definitely, I've already created issue for it https://github.com/kiwicom/universal-components/issues/201

src/Picker/Picker.ios.js
JoseDuSV: Whole duplicate native picker implementation extracted to separate component and duplicate function added to `PickerHelpers`.

src/Picker/Picker.ios.js
JoseDuSV: Label extracted to props as in other cases.

src/Picker/Picker.web.js
JoseDuSV: Sure. Short description added to const. Hope it will be clearer now.

JoseDuSV:
iOS version now uses modal to show options. So the implementation is more flexible than in first iteration.

picker-ios

src/Picker/__tests__/index.test.js
```diff
@@ -0,0 +1,38 @@
...
+      onValueChange={onValueChange}
+      confirmLabel={confirmLabel}
+    />,
+  );
+
+  it('should contain correct label', () => {
+    expect(getByText(testData[1].label)).toBeDefined();
+  });
+
+  it('renders', () => {
```
RobinCsl: Sometimes I wonder of the usefulness and value of these shallow snapshots. Maybe the previous test is enough to assert that your component at least rendered the right label.

PR #206, opened 2019-01-23
Fix: orbit-icons font added to built documentation

RobinCsl:
Summary: It is somehow tricky to add custom fonts in Docz, and a few config properties are not fully documented. I managed to make the fonts work only in the built result, not in the dev environment. So this could be improved in the future.


RobinCsl:
Preview of docs
https://deploy-preview-206--kiwicom-universal-components-docs.netlify.com/icon

JoseDuSV:
> Preview of docs
> https://deploy-preview-206--kiwicom-universal-components-docs.netlify.com/icon

We will probably need to set redirect rules ( https://www.netlify.com/docs/redirects/ ), because for now "subpages" links like `/icon` doesn't work. Is it easy to fix or should I create separate issue for it and we will also polish this later?

RobinCsl:
@JoseDuSV It should be fixed now 😄

JoseDuSV:
One small thing. I'm getting flow error before I run `docz:dev` for the first time.

screenshot 2019-01-23 at 15 08 54

Maybe we can disable `flow` for `doczrc.js`.

RobinCsl:
I think this would be only if you run `yarn docz:dev` from the `docs` folder, without having done `yarn install` before. But if you do it from the root of the repo, it should be fine I believe.

JoseDuSV:
> I think this would be only if you run `yarn docz:dev` from the `docs` folder, without having done `yarn install` before. But if you do it from the root of the repo, it should be fine I believe.

Maybe I'm doing something wrong. 
But if I try to reproduce it:
* Delete installed docs dependencies `docs/node_modules` 
* Run `yarn install` from root
* Run `yarn flow`

Mentioned `flow` error is still there.

I need to run run `docz:dev` or `docz:build` from root (simply command which contains `cd docs; yarn;`) to install required docs dependencies.

RobinCsl:
Exactly, that's because we don't have Yarn workspaces in this repo. I know that I had to move Docz to its own folder as it was conflicting with Storybook's version of emotion. So right now, it is necessary to install dependencies in `docs/`. It is automatic if you do `yarn docz:dev`, it's not if you do `yarn` from root. 
Makes more sense?

PR #208, opened 2019-01-25
Feat: Added visual regression testing

RobinCsl:
Summary: Like on `orbit-components`, we use [Loki](https://loki.js.org) to perform visual testing.
Instructions to run the tests were added to CONTRIBUTING, as well as how to let Loki skip stories which are too dependent on context for them to always look the same. A good example is the Playground for the DatePicker component; a new date is generated at each render.
This PR also adds a .loki folder at the root, containing screenshots of all the (non-skipped) stories we wrote.


FilipMessa:
I just realize that there are the missing references for `chrome_laptop_*`, because when I started the `loki:approve:chrome` command it generated it all files for me

JoseDuSV:
As Filip mentioned, just add the missing `chrome_laptop_*` references. And otherwise it's perfect.

PR #211, opened 2019-01-29
Feat: Added Modal component

JoseDuSV:
Summary: Extracted `Modal` component from `Margarita`, because it is used even by some `universal-components` like `DatePicker` or `Picker`.

Implementation was updated to make `Modal` more flexible.
Native version was mostly replaced by implementation from `mobile` repo: https://github.com/kiwicom/mobile/blob/master/packages/shared/src/Modal.js .

Thanks @tbergq for pointing me to the current `mobile` implementation.


src/DatePicker/DatePicker.ios.js
```diff
@@ -71,55 +67,43 @@ export default class iOSDatePicker extends React.Component<Props, State> {
...
+            mode={mode}
+            maximumDate={maxDate}
+            minimumDate={minDate}
+          />
+          <View style={styles.buttonsContainer}>
+            <Touchable
+              onPress={this.handleDismiss}
+              style={styles.confirmButton}
+            >
+              <Text style={styles.buttonText}>Cancel</Text>
```
tbergq: This should probably be passed in from the consumer. Since we would want to translate it.

src/DatePicker/DatePicker.ios.js
```diff
@@ -71,55 +67,43 @@ export default class iOSDatePicker extends React.Component<Props, State> {
...
+            </Touchable>
+            <Touchable
+              style={styles.confirmButton}
+              onPress={this.handleConfirm}
+            >
+              <Text style={styles.buttonText}>OK</Text>
```
tbergq: Same here

src/Modal/Modal.web.js
```diff
@@ -0,0 +1,60 @@
...
+}: Props) => {
+  const dynamicStyle = StyleSheet.create({
+    backdrop: {
+      opacity: backdropOpacity ?? 0.5,
+      backgroundColor: backdropColor ?? defaultTokens.paletteInkDark,
+    },
+  });
+
+  return (
+    <>
```
tbergq: is this fragment necessary?

src/Modal/ModalTypes.js
```diff
@@ -0,0 +1,24 @@
+// @flow
+
+import * as React from 'react';
+import type { StylePropType } from '../PlatformStyleSheet/StyleTypes';
+
+/**
+ * NOTE:
+ * Props are not exact - additional properties allowed.
+ * Available props for native implementation: https://github.com/react-native-community/react-native-modal#available-props
```
tbergq: But props are exact 😄 you can delete this comment

src/DatePicker/DatePicker.ios.js
RobinCsl: One way we could achieve this is by adding additional prop:
```
type Props = {|
...
+label: {|
   +cancel: string,
   +confirm: string,
|},
...
|};
```

tbergq: But please do not restrict it to string, it could also be any component, `Translation` or `Text`

RobinCsl: So replace `string` with `React.Node`?

tbergq: That should be good enough yes 😊

JoseDuSV: Updated

src/DatePicker/DatePicker.ios.js
JoseDuSV: Updated

JoseDuSV:
> Any chance to have visual regression testing screenshots added? @JoseDuSV
> (If it's not relevant, can you please modify the story not with `lokiSkip` instead of `add` 😉 )

Sure, thanks for reminder. At the end I've added `lokiSkip` in this case because it was just "Open Modal" button (and it is not ideal to have modal visible as default because of mobile).

RobinCsl:
@JoseDuSV Well, you could always add a prop with default value `false`. That's sort of what I have done for DatePicker if I remember correctly.
Also, regarding the appearance of the modal, is the design to have very angular/rectangular borders? This could work on Android, but on iOS it looks unusual.

JoseDuSV:
> @JoseDuSV Well, you could always add a prop with default value `false`. That's sort of what I have done for DatePicker if I remember correctly.
> Also, regarding the appearance of the modal, is the design to have very angular/rectangular borders? This could work on Android, but on iOS it looks unusual.

@RobinCsl Right, I've used same playground structure as for `DatePicker` and added modal screenshots. 
Plus also updated affected `DatePicker` screenshots (modal background opacity and cancel button text was changed).

Regarding the background, white rectangle which you can see in playground is actually not part of the modal anymore. The goal of this PR was also to make modal more flexible. So it is basically just black overlay above other content and you can pass anything as children. So it can be now used for (default) centered content like in case of `DatePicker` or bottom-aligned components like for example pickers in Margarita or new version of `Picker`.

screenshot 2019-01-31 at 16 07 19

JoseDuSV:
@RobinCsl Just updated playgrounds for `DatePicker` and `Modal` to make them closable on mobile and created new screenshots. I also swapped default dates in `DatePicker` playground to fixed, to be sure that date on screenshots will not change every time Loki is launched.

PR #214, opened 2019-01-30
Fix: fixed the RowOption arrow design on android

FilipMessa:
**Before:**
screen shot 2019-01-30 at 13 53 24

**After**
screen shot 2019-01-30 at 14 09 54


FilipMessa:
close #214

PR #219, opened 2019-02-04
Chore: Upgrading babel packages and rimraf

RobinCsl:


RobinCsl:
#207 seems to have problems in the deployed storybook. So I will upgrade one by one.

PR #243, opened 2019-02-13
Refactor: change the OptionPicker from uncontrolled to controlled component

FilipMessa:
I had to refactor the OptionalPicker from uncontrolled component to controlled component. 
As uncontrolled and stateful components it works nice but just in the storybook, If I tried implemented this in the Margarita I realized, that I need
changes the OptionalPicker's state from parents, there were two options one
use ' getDerivedStateFromProps()' but this will lead to a lot of complexity.
So I decided to change OptionalPicker to a stateless uncontrolled component.

### DIFF
--- a/src/OptionPicker/OptionPicker.js
+++ b/src/OptionPicker/OptionPicker.js
@@ -8,142 +8,61 @@ import type { OptionType } from './OptionPickerTypes';
 import { StyleSheet } from '../PlatformStyleSheet';
 import { TagsInput } from '../TagsInput';
 import OptionList from './components/OptionList';
-import { getOptionsIds, areSelectedOptionsChanged } from './helpers';
 
 type Props = {|
+  +onChangeText: string => void,
+  +onClearPress: () => void,
+  +onPressAdd: OptionType => void,
+  +onPressItem: OptionType => void,
+  +text?: string,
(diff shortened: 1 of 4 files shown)


src/OptionPicker/OptionPicker.js
```diff
@@ -8,142 +8,61 @@ import type { OptionType } from './OptionPickerTypes';
...
+          placeholder={placeholder}
+        />
       </View>
-    );
-  }
+      {options && (
+        <OptionList
+          onItemPress={onPressItem}
+          onAddPress={onPressAdd}
+          options={options}
```
RobinCsl: The logic changed here for your filteredOptions.

FilipMessa: I know about that, I was thinking about and discussed this also with @JoseDuSV.
And the idea behind is this filtering shoul be in parent component, because sometime maby you want to have the option to choose one option more that one time. 
But I have not a strong opinion about that I can return it if you prefer to keep it here.

RobinCsl: Okay, let's just iterate on this. I'll accept it as it is for now because the placepicker PR has been waiting for too long, and it'd be nice to have it ready tomorrow.

PR #252, opened 2019-02-22
Styling prop for TextInput's label container

jvaclavik:
Without this, I'm not able to style form labels properly, because I can change only text-related styles

screenshot 2019-02-25 at 10 10 52

### DIFF
--- a/src/TextInput/FormLabel.js
+++ b/src/TextInput/FormLabel.js
@@ -5,7 +5,7 @@ import { defaultTokens } from '@kiwicom/orbit-design-tokens';
 import { View } from 'react-native';
 
 import { Text } from '../Text';
-import { StyleSheet } from '../PlatformStyleSheet';
+import { StyleSheet, type StylePropType } from '../PlatformStyleSheet';
 
 type Props = {|
   +children: React.Node,
@@ -13,6 +13,7 @@ type Props = {|
   +disabled?: boolean,
   +required?: boolean,
   +inlineLabel?: boolean,
+  +style?: StylePropType,
 |};
 
 const getAsteriksStyle = (filled, disabled) => {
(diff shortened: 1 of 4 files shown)


JoseDuSV:
Can you please just shortly update description why this is necessary. 

As mentioned in #253 there is also inline version of label so it should probably also be affected by this prop.

https://github.com/kiwicom/universal-components/blob/79fab54b8e547fda698016c576cae8ebad1f1858/src/TextInput/TextInput.js#L213-L222

jvaclavik:
> Can you please just shortly update description why this is necessary.

Done

> As mentioned in #253 there is also inline version of label so it should probably also be affected by this prop.

Good point! I didn't know about inline labels, now it should be fixed.

jvaclavik:
> Can you please just shortly update description why this is necessary.
> 
> As mentioned in #253 there is also inline version of label so it should probably also be affected by this prop.
> 
> [universal-components/src/TextInput/TextInput.js](https://github.com/kiwicom/universal-components/blob/79fab54b8e547fda698016c576cae8ebad1f1858/src/TextInput/TextInput.js#L213-L222)
> 
> Lines 213 to 222 in [79fab54](/kiwicom/universal-components/commit/79fab54b8e547fda698016c576cae8ebad1f1858)
> 
>  <InlineLabel> 
>    <FormLabel 
>      filled={!!value} 
>      inlineLabel 
>      required={required} 
>      disabled={disabled} 
>    > 
>      {label} 
>    </FormLabel> 
>  </InlineLabel>

Done

> Thanks. Looks good.
> 
> @jvaclavik Maybe if you need to add `marginLeft` to all form text inputs we can consider setting this value as default one.
> This component was one of the first, so it is possible that it wasn't styled according to final designs.

OK, we can do it. And I would keep possibility of changing View's style.

JoseDuSV:
@jvaclavik 👍 Definitely keep styling possibility, it can be handy later.

jvaclavik:
> @jvaclavik 👍 Definitely keep styling possibility, it can be handy later.

`marginLeft` added to `TextInput` for block-styled labels

PR #253, opened 2019-02-22
[discussion needed] Separate FormLabel from TextInput

jvaclavik:
Labels should not be depend only to `TextInput`. For example we need it also in `Picker` so I think it should be exportable from `universal-components`. 

## 1. Old form
```
<TextInput label="Address" ... />
```

## 2. New form
```
<FormLabel label="Address" />
<TextInput ... />
```

But there is a problem with duplicity, because we still can do something like that:
```
<FormLabel label="Address" />
<TextInput label="Address" ... />
```

So I'm thinking about deleting `label` prop from `TextInput`. Are you OK with that?

I also noticed that there could be a problem with using `flexDirection: "row"`, because it affects the position of the label. For example:

```
<View style={{flexDirection: 'row'}}>
  <FormLabel label="Address" />
  <TextInput label="Address" ... />
</View>
```

So that is the reason why I think we should use another form:

## 3. Proposed form

In this form you are not affected by using parent's `flexDirection`. It is also more logical in semantic way but there is disadvantage that you have to set another parameter for replace label's `Text` with custom component. So we would have to add new prop like `labelComponent`.

```
<FormLabel label="Address">
  <TextInput ... />
</>
```

or

```
<FormLabel labelComponent={<View style={styles.label}><Text>Address</Text></View>}>
  <TextInput ... />
</>
```

What do you think about that? 

@JoseDuSV @FilipMessa @RobinCsl


JoseDuSV:
And what about inline version of label for compact input?
I'm worried that it needs to stay as part of the component.
screenshot 2019-02-22 at 15 53 37

I totally understand your motivation. I've been thinking about something similar recently with input error message. But in this case I'm slightly worried if it doesn't adding more complexity than it solves.

What about just exporting `FormLabel` from `universal-components` and implementing it directly to mentioned picker component?

jvaclavik:
> And what about inline version of label for compact input?
> I'm worried that it needs to stay as part of the component.
> screenshot 2019-02-22 at 15 53 37

In this case `label` prop for `TextInput` could totally make sense for me.


> I totally understand your motivation. I've been thinking about something similar recently with input error message. But in this case I'm slightly worried if it doesn't adding more complexity than it solves.
> 
> What about just exporting `FormLabel` from `universal-components` and implementing it directly to mentioned picker component?

This is also good solution. But for example, you can't use label which you pasted above.

PR #254, opened 2019-02-25
Add FormLabel to Picker

jvaclavik:


src/Picker/Picker.android.js
```diff
@@ -25,15 +28,20 @@ const Picker = ({
...
-      <Icon name={iconName ?? 'chevron-right'} style={styles.icon} />
-      <NativePicker
-        optionsData={optionsData}
-        selectedValue={selectedValue}
-        style={styles.picker}
-        onValueChange={onValueChange}
-      />
+    <View>
+      {label != null && (
+        <FormLabel style={labelContainerStyle}>{label}</FormLabel>
```
JoseDuSV: We should probably add initial style with `marginStart` like in case of `TextInput` to make it unified.
Maybe it will be even better to set this style directly inside `FormLabel` if we want it unified by default in all implementations.

jvaclavik: Makes sense, added `marginStart` to `FormLabel`

